PR #1, opened 2020-10-05
Master

emmastephan:
Added in bottom tab navigation with icons that change color when clicked on, and a header with a hamburger icon that opens a menu that slides out from the side.


jenmhoang:
closing and moving PR to the properly named branch

PR #2, opened 2020-10-17
Abstracting Airtable Requests

mdterrenal:
## 🖋 PR Description

This PR creates abstractions of API requests to Airtable within requests.ts and also defines the schema and interface for our data based on the database.

### 🐞 Bug Report

- There were lots of bugs related to converting everything to the correct Record types, but I figured them out by searching through past codebases and referencing their methods.

### 🧪 Testing

How do we test your feature?

1. First you need to set up your .env file containing all the environment variables that you need to make requests to Airtable. Create a file in your directory called .env, and copy the contents of .env.example (that should be in this PR) into that file. REACT_APP_AIRTABLE_BASE_ID should be set equal to the ID of our base, which you can find here: https://airtable.com/appy54OSlg5bKS8rU/api/docs. REACT_APP_AIRTABLE_API_KEY should be set equal to your Airtable API key, which you can set up by going to API under your Account overview/settings on Airtable. BE SURE THAT YOU DO NOT COMMIT THIS FILE!! .gitignore should already prevent this from happening, but just verify that you don't accidentally commit it.

2. Once that's done, you can follow the example below to see how you can pull data from Airtable!
backendpullexample

3. You should be able to use npm start to run the code and see your results!

cc: @jenmhoang


jenmhoang (approved):
it took me a while to read through and test all of this but this is -chef's kiss- thank you for beautifully tackling such a hefty task

src/utils/airtable/interface.ts
```diff
@@ -0,0 +1,183 @@
...
+}
+
+export interface UserRecord extends Record {
+  အမည်: string;
+  name: string;
+  username: string;
+  email: string;
+  password: string;
+  photo?: Airtable.Attachment[];
+  incidentIds: string[];
```
julianrkung, line 34: Should some of these properties in these interfaces be optional (ex: Users aren't required to have an email)? It might make sense to establish the optional-ness later, though. It seems like a lot of invariants to enforce right now before we understand how our data really works.

julianrkung (approved):
u r gr8. lgtm w/a comment

julianrkung:
Also is there a plan on how this typescript stuff is going to be maintained on future schema regenerations?

julianrkung:
@mdterrenal can this get merged if we don't have any other changes to make? I want to merge my airlock PR also but I want to change the airtable-schema-generator to use airlock before my merge

src/utils/airtable/interface.ts
mdterrenal: yeah, i didn't want to define optional-ness for certain properties because i wasn't 100% sure what would/wouldn't be necessary given how the data works but definitely something to adjust later on!

PR #5, opened 2020-10-18
Add offline service worker

julianrkung:
PR contains the following

1. Custom service worker that records POST requests (right now hardcoded to a localhost)
2. A test component placed on the homepage that sends a POST request to a hardcoded backend
3. Overwrote `npm run build` to use new custom service worker and clean old service worker.

Relevant customer service worker information:
- Worker set for cache first strategy (we can change this if that's a problem. Might want network-first).
- Worker currently caches fonts, images, static pages in user's indexedb.
- Worker caches failed post requests (to be sent on background-sync) in indexedb under object "PostQueue"


jenmhoang (approved):
lgtm! We can also think about how a user can be notified/prompted to reload/update in the future (can discuss with Ashley)

PR #8, opened 2020-10-23
Add Bottom Navigation

jenmhoang:
original PR by @emmastephan 
> Added in bottom tab navigation with icons that change color when clicked on, and a header with a hamburger icon that opens a menu that slides out from the side.


emmastephan (commented):
LGTM!!

PR #9, opened 2020-10-23
Kyle/customers

kyleqhua:
🖋 PR Description
Customer page. Can search through records. Sort/Filter feature not implemented because need to look at Airtable date. Airtable integration in progress. Need a little css tweaks.

🤳Screenshots (if applicable)
ezgif-4-8176b295554a

🐞 Bug Report

🧪 Testing


julianrkung:
Is this blocked on anything?

src/components/CustomerCard.tsx
```diff
@@ -0,0 +1,101 @@
...
+    singleCard: string; titleText: string; totalText: string;
+    numberText: string; divSpacing: string
+  };
+}
+
+function CustomerCard(props: CardProps) {
+  const { classes } = props;
+  return (
+    <Card className={classes.singleCard}>
+      <div className={classes.cardCon}>
```
julianrkung: since `cardContent` is decently short, lets not abbreviate it so it's clearer. Took me a little bit to figure out what it meant.

src/components/HamburgerIcon.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const Hamburger = (props: Props) => (
+  <StyledHamburger open={props.open} onClick={() => props.setOpen(!props.open)}>
+    <div />
+    <div />
+    <div />
+  </StyledHamburger>
+);
+
+export default Hamburger;
```
julianrkung: Is this component used? I think it was deleted some time ago, so if you're not using it can you delete this file?

src/components/NavBar.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
```
julianrkung: Is this component used? I think it was deleted some time ago and replaced with `BottomNavigation`. Can you check that it's no longer used in the most current version and delete it if it isn't?

src/components/Navigation.tsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
```
julianrkung: Is this component used? I think it was deleted some time ago and replaced with BottomNavigation. Can you check that it's no longer used in the most current version and delete it if it isn't?

src/components/NavigationBar.js
```diff
@@ -0,0 +1,27 @@
+import React, { Component } from 'react'
```
julianrkung: Is this component used? I think it was deleted some time ago and replaced with BottomNavigation. Can you check that it's no longer used in the most current version and delete it if it isn't?

src/styles/HamburgerStyle.ts
```diff
@@ -0,0 +1,55 @@
+import styled from 'styled-components';
```
julianrkung: I don't think this file is used anymore since Hamburger isn't. Can you double check and delete if its not?

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,130 @@
...
+  const calcLatestReadingDate = (curr: CustomerRecord) => {
+    //depends if the meter readings list is sorted with earliest => latest
+    let latestMeterReading = curr.meterReadings[curr.meterReadings.length - 1]
+    const dateTime = Date.parse(latestMeterReading.date)
+    let readingDate = new Date(dateTime)
+    let month = readingDate.getMonth() + 1;
+    let day = readingDate.getDate();
+    let shortDate = month.toString() + '.' + day.toString();
+    // console.log(readingDate.toLocaleDateString())
+    // console.log(shortDate)
```
julianrkung: Delete commented out code

src/components/hooks.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, RefObject } from 'react';
```
julianrkung: Is this used anywhere? I grep'd for it and i can't find any uses. Delete file/function if not used.

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,130 @@
...
-    setCustomers(customers);
-  };
+  const getCustomers = () => {
+    const siteData = store.getState().siteData;
+    const sites: SiteRecord[] = siteData.sites;
+    let allCustomers: CustomerRecord[] = [];
+    for (let i = 0; i < sites.length; i++) {
+      let currCustomers: CustomerRecord[] = sites[i].customers;
+      allCustomers = allCustomers.concat(currCustomers)
+    }
```
julianrkung: Doesn't this grab the customers from all the sites the user has access to? It's my understanding that we only want the customers of the current site the user has set.

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,130 @@
...
+      <div className={classes.headerWrapper}>
+        <h1 className={classes.title}>Customers</h1>
+        <div className={classes.selectionHeader}>
+          <UserSearchBar searchFun={handleSearchChange} />
+          <FormControl>
+            <Select inputProps={{ 'aria-label': 'Without label' }}>
+              {/* placeholder sorting for now */}
+              <MenuItem value={10}>Ten</MenuItem>
+              <MenuItem value={20}>Twenty</MenuItem>
+              <MenuItem value={30}>Thirty</MenuItem>
```
julianrkung, line 97: Can we get this `Sort By` working in this PR? Is there a reason it's placeholdered / on backburner?

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,130 @@
...
+    console.log(allCustomers)
+    const customersTrie: TrieTree<CustomerRecord> = new TrieTree('name');
+    customersTrie.addAll(allCustomers);
+    setAllCustomersTrie(customersTrie);
 
+  }
+
+  const calcLatestReadingDate = (curr: CustomerRecord) => {
+    //depends if the meter readings list is sorted with earliest => latest
+    let latestMeterReading = curr.meterReadings[curr.meterReadings.length - 1]
```
julianrkung: We need to account for empty array case if customer has no meter readings

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,130 @@
...
+    setCustomers(allCustomers);
+    setFullCustomers(allCustomers);
+    console.log(allCustomers)
+    const customersTrie: TrieTree<CustomerRecord> = new TrieTree('name');
+    customersTrie.addAll(allCustomers);
+    setAllCustomersTrie(customersTrie);
 
+  }
+
+  const calcLatestReadingDate = (curr: CustomerRecord) => {
```
julianrkung: `customer` instead of `curr`. I also prefer `getLatestMeterReadingDate` to the current name since nothing is being calculated but that's small

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,130 @@
...
+
+function CustomerMain(props: RouteComponentProps & UserProps) {
   useEffect(() => {
     getCustomers();
   }, []);
 
+  const { classes } = props;
   const [customers, setCustomers] = useState<CustomerRecord[]>([]);
+  const [allCustomersTrie, setAllCustomersTrie] = useState<TrieTree<CustomerRecord>>(new TrieTree('name'));
+  const [fullCustomers, setFullCustomers] = useState<CustomerRecord[]>([]);
```
julianrkung, line 47: Is this `fullCustomers` used anywhere? We might be able to get away with just `customers` (maybe `filteredCustomers` or some indication that this is the customers being shown is a better name), and the `allCustomersTrie` which will always hold all the customers afaik

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,130 @@
...
+  });
+
+interface UserProps {
+  classes: { title: string; headerWrapper: string; selectionHeader: string; scrollDiv: string };
+}
+
+function CustomerMain(props: RouteComponentProps & UserProps) {
   useEffect(() => {
     getCustomers();
   }, []);
```
julianrkung, line 54: Performance might suffer here since we call `getCustomers` (which iterates through all customers and does some sets) every time we do a rerender, which happens when we search. We might just want to call this `getCustomers` on the initial render to save all the extra work when we render during filters. Open to discuss on this though if you have opinions

julianrkung (changes requested):
Can you delete the empty/outdated files in this PR?

Left some comments on code logic first. When you do that push for green icons I'll take another look and also comment on frontend (which should just be small changes).

src/components/HamburgerIcon.tsx
kyleqhua: yeah i think its left over from old master

src/screens/Customers/CustomerMain.tsx
kyleqhua: wait yeah that prob more apporpriate in regards to the app.

src/screens/Customers/CustomerMain.tsx
kyleqhua: I think Jen originally gave the task to Emma? So I didn't implement it. I can do it if needed.

src/screens/Customers/CustomerMain.tsx
kyleqhua: I forgot that the trie holds all the customers. Thanks!

src/components/CustomerCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+      fontStyle: 'normal',
+      fontWeight: 500,
+      fontSize: '12px',
+      color: '#BDBDBD',
+    },
+    totalText: {
+      fontFamily: 'Helvetica Neue',
+      fontStyle: 'normal',
+      fontWeight: 500,
+      fontSize: '12px',
```
julianrkung: Replace with `body1` in `ThemeStyles.ts` so if it gets changed we don't have to change all individually

src/components/CustomerCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+      fontSize: '20px',
+      lineHeight: '98.1%',
+      color: '#828282',
+      paddingTop: '14px'
+    },
+    updatedText: {
+      fontFamily: 'Helvetica Neue',
+      fontStyle: 'normal',
+      fontWeight: 500,
+      fontSize: '12px',
```
julianrkung: Replace with `body1` in `ThemeStyles.ts` so if it gets changed we don't have to change all individually

src/components/UserSearchBar.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+import { Button, TextField, Typography, Link, InputBase, Paper, createStyles, Theme, withStyles } from '@material-ui/core';
+import SearchIcon from '@material-ui/icons/Search';
+interface SearchProps {
+  searchFun: any;
```
julianrkung: something related to `onSearchChange` might be more appropriate since `searchFun` isn't as descriptive as it should be imo. Try not to abbreviate into words like `fun` because it's harder to understand immediately what this is. Clarity in variable names is gonna be really important if NPO is going to be doing development after handoff

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,120 @@
...
+  createStyles({
+    title: {
+      width: '153.9px',
+      height: '33.83px',
+      left: '25.6px',
+      top: '96px',
+      fontFamily: 'Helvetica Neue',
+      fontStyle: 'normal',
+      fontWeight: 500,
+      fontSize: '30px',
```
julianrkung: Replace with `h1` in `ThemeStyles.ts` so if it gets changed we don't have to change all individually

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,120 @@
...
+  const [filteredCustomers, setFilteredCustomers] = useState<CustomerRecord[]>([]);
+  const [allCustomersTrie, setAllCustomersTrie] = useState<TrieTree<CustomerRecord>>(new TrieTree('name'));
 
-  const getCustomers = async () => {
-    const customers: CustomerRecord[] = await getAllCustomers();
-    setCustomers(customers);
-  };
+  const getCustomers = () => {
+    const siteData = store.getState().siteData.currentSite;
+    let allCustomers: CustomerRecord[] = siteData.customers;
```
julianrkung: use `const` if the variable isn't gonna change

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,120 @@
...
+
+  const getLatestReadingDate = (customer: CustomerRecord) => {
+    //depends if the meter readings list is sorted with earliest => latest
+    if (customer.meterReadings.length > 0) {
+      let latestMeterReading = customer.meterReadings[customer.meterReadings.length - 1]
+      const dateTime = Date.parse(latestMeterReading.date)
+      let readingDate = new Date(dateTime)
+      let month = readingDate.getMonth() + 1;
+      let day = readingDate.getDate();
+      let shortDate = month.toString() + '.' + day.toString();
```
julianrkung: use `const` if the variable isn't gonna change

src/screens/Customers/CustomerMain.tsx
julianrkung: Talked with Kyle offline. This'll be done in a subsequent PR.

julianrkung (changes requested):
Looking good, probably just like ~1-2 more iterations!

Could you delete all the empty, outdated files too (look at PR diff to find them all)?

Also I found a bug. If you search on the page and then delete the search (so the search bar is empty, you end up with no customers. Looks like a change needs to be made to `handleSearchChange`
image


Last thing: The screen looks a little off on different phones (especially something like the `iPhone 6/7/8 Plus`. I think this has to do with the hardcoded pixel styling that's being used. I don't think we can trust/copy-paste the Figma-generated CSS for stuff like widths and margins etc. @jenmhoang did a good job building base components to use so that our screens look uniform. Look at how `CustomerProfile` and `Home` are structured (using `BaseScreen` etc...) to make sure that the page looks good on all platforms.

(this is a screenshot of how it looks on iphone plus sizes)
image

src/components/UserSearchBar.tsx
kyleqhua: yessir that makes sense.

src/screens/Customers/CustomerMain.tsx
kyleqhua: I remember why we needed full customers. There is no way to get the full customers from the trie. With an empty string '' the trie returns nothing. Looked fpr documentation and it seems like you cant get it all back. I looked up how the trie was used in Unloop and we used the same method (keep a full copy of list): 
https://github.com/calblueprint/unloop/blob/10b27cbc70fbc85d2322db00d3fc7088facd2713/app/javascript/components/StaffDashboard/index.js#L45

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,124 @@
...
+import BaseScrollView from '../../components/BaseComponents/BaseScrollView';
+import TrieTree from '../../lib/utils/TrieTree';
+import BaseScreen from '../../components/BaseComponents/BaseScreen';
 
-function CustomerMain(props: RouteComponentProps) {
+const styles = (theme: Theme) =>
+  createStyles({
+    title: {
+      width: '153.9px',
+      height: '33.83px',
```
julianrkung: This causes awkward spacing between the search bar and title. See screenshots.

With:
image

Without:
image

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,124 @@
...
+      fontWeight: theme.typography.h1.fontWeight,
+      fontSize: theme.typography.h1.fontSize,
+      lineHeight: '115%',
+      color: '#828282',
+      flexGrow: 1,
+      paddingRight: '20px',
+    },
+    headerWrapper: {
+      display: 'flex',
+      flexDirection: 'row',
```
julianrkung, line 26: i think this is default so there's no need to put `flexDirection: row`

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,124 @@
...
 
   return (
-    <>
-      {customers.map((customer, index) => (
-        <Link key={index} to={{ pathname: `${props.match.url}/${customer.name}`, state: { customer: customer } }}>
-          <h3>{customer.name}</h3>
-        </Link>
-      ))}
-    </>
+    <BaseScreen>
```
julianrkung: Add `rightIcon user` to match Figma.

Right now:
image

Figma:
image

src/components/CustomerCard.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+const styles = (theme: Theme) =>
+  createStyles({
+    arrow: {
+      color: theme.palette.primary.main,
+    },
+    cardContent: {
+      flexGrow: 1,
+      display: 'flex',
+      flexDirection: 'row',
```
julianrkung: default flexDirection is row. can be removed

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,29 +1,124 @@
 import React, { useEffect, useState } from 'react';
-import { CustomerRecord } from '../../lib/airtable/interface';
-import { getAllCustomers } from '../../lib/airtable/request';
+import { CustomerRecord, SiteRecord } from '../../lib/airtable/interface';
 import { Link, RouteComponentProps } from 'react-router-dom';
+import { store } from '../../lib/redux/store';
+import { createStyles, FormControl, FormHelperText, MenuItem, Select, Theme, ThemeProvider, withStyles } from '@material-ui/core';
+import UserSearchBar from '../../components/UserSearchBar';
+import CustomerCard from '../../components/CustomerCard';
+import BaseScrollView from '../../components/BaseComponents/BaseScrollView';
```
julianrkung: Component isn't being used so we can delete

src/components/CustomerCard.tsx
```diff
@@ -0,0 +1,123 @@
+import React from 'react';
+import { createStyles, makeStyles, Theme, withStyles } from '@material-ui/core/styles';
+import { Card, CardActions, CardContent, Button, Typography, IconButton, StylesProvider } from '@material-ui/core';
```
julianrkung: Can you delete the unused imports here?

src/components/UserSearchBar.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+import { Button, TextField, Typography, Link, InputBase, Paper, createStyles, Theme, withStyles } from '@material-ui/core';
+import SearchIcon from '@material-ui/icons/Search';
```
julianrkung: Can you delete the unused imports here?

src/components/UserSearchBar.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const styles = (theme: Theme) =>
+  createStyles({
+    search: {
+      color: '#828282',
+    },
+    searchWrapper: {
+      width: '100%',
+      height: '25px',
+      border: '1px solid #ff922d',
```
julianrkung, line 30: According to Figma, border should be the `main` palette in `ThemeStyles.ts`  when active and otherwise `BDBDBD`. Lets put off the searchbar animation for another PR if we decide to do it since it's not that important.

Figma screenshot:
image

src/components/UserSearchBar.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+import { Button, TextField, Typography, Link, InputBase, Paper, createStyles, Theme, withStyles } from '@material-ui/core';
+import SearchIcon from '@material-ui/icons/Search';
+interface SearchProps {
+  onSearchChange: any;
+  classes: { search: string; searchWrapper: string; inputWrapper: string; };
+}
+
+const styles = (theme: Theme) =>
```
julianrkung: If the styles don't use theme, you don't need to have it as a parameter. In general, don't add a parameter unless it's going to be used (this might change since we should be using the `primary.main` from themeStyles.ts).

src/components/UserSearchBar.tsx
```diff
@@ -0,0 +1,39 @@
...
+  createStyles({
+    search: {
+      color: '#828282',
```
julianrkung, line 13: The search icon looked like it was a little offcenter from the search. To regulate you can restrict the dimensions to be proportional to searchWrapper. I put a `height: 75%` and it looks nicer. You can play around and see if there's anything better but feel free to use 75% otherwise.

Current:
image

Changed icon dimension:
image

src/screens/Customers/CustomerMain.tsx
julianrkung: Sounds good, thanks for checking it! :^)

julianrkung (changes requested):
Final comments. Should be good to merge after all these are addressed.

One last thing, it looks like scroll doesn't behave how it's supposed to on the page. I believe the top title / search bar / sort should be sticky on scroll. That's how it is on the `CustomerProfile.tsx` at least. Attached a couple clips to show what I mean.

Current / nonsticky header:
https://user-images.githubusercontent.com/35501399/104109947-b2471e80-5287-11eb-9254-9c6d275a4df4.mov

Sticky header / customer profile:
https://user-images.githubusercontent.com/35501399/104109950-bd9a4a00-5287-11eb-83c5-d6f112988225.mov

julianrkung (approved):
Thanks for working so hard on this! :^)

PR #10, opened 2020-11-01
Integrate Airlock into frontend. Add functionality to register and login users.

julianrkung:
Airlock is used to register and log in users in the login page.

Testing PR requires a working Airlock server as well as a clientside `.env` file to run.

PR is WIP because it does not save user data yet (will use redux for that).

Screenshots:
image
image


src/lib/redux/userData.ts
```diff
@@ -0,0 +1,10 @@
+import { store } from './store';
+import { saveUserData, setLoadingForUserData } from './userDataSlice';
+
+// TODO: Change from ANY type to user schema after Maria's schema PR gets merged
```
jenmhoang: Maria's PR has since been merged if you want to update this!

jenmhoang (approved):
lgtm! 🚀

mdterrenal (approved):
everything looks good to me as well!

src/lib/redux/userData.ts
julianrkung: Looks like the typing isn't as easy as I thought since the fields surprisingly don't match :( . I updated the TODO and will do it another time!

PR #11, opened 2020-11-02
Customer Profile screen with navigation

jenmhoang:
🖋 PR Description
- Finished front-end for the main Customer Profile screen, built from base components (Base Header, OutlinedCardList)
- implemented MUITheme and withStyles styling conventions
- added navigation that sends the appropriate records to the next screen
- pulled records from Airtable

🤳Screenshots (if applicable)
ezgif-7-de01e5ec12bd


src/screens/Customers/CustomerMain.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  const getCustomers = async () => {
+    const customers: CustomerRecord[] = await getAllCustomers();
+    setCustomers(customers);
+  };
+
+  return (
+    <>
+      {customers.map((customer, index) => (
+        <Link key={index} to={{ pathname: `${props.match.url}/${customer.name}`, state: { customer: customer } }}>
```
julianrkung: URL uniqueness from previous comment

src/App.tsx
```diff
@@ -21,11 +22,13 @@ function App() {
...
             <Switch>
               <Route path="/home" component={Home} />
-              <Route path="/customers" component={User} />
+              <Route path="/customers" exact component={CustomerMain} />
               <Route path="/inventory" component={Inventory} />
               <Route path="/maintenance" component={Maintenance} />
               <Route path="/incidents" component={Incidents} />
               <Route path="/login" component={Login} />
+              <Route path={'/customers/:name'} exact component={CustomerProfile} />
+              <Route path={'/customers/:name/records'} component={CustomerRecords} />
```
julianrkung: Wouldn't we get URL collisions if customers had the same name? We might want to use user ID as the endpoint here to ensure uniqueness.

src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -0,0 +1,68 @@
...
+  title?: string;
+  rightIcon?: string;
+  classes: any;
+}
+
+function BaseHeader(props: HeaderProps) {
+  const { classes } = props;
+  const history = useHistory();
+
+  const getIcon = (onClick: any, icon: JSX.Element) => {
```
julianrkung: `any` usage. Would it work if we replace it with an `() => void` or `(event : React.MouseEvent) => void` type?

src/components/OutlinedCardList.tsx
```diff
@@ -0,0 +1,74 @@
...
+      </div>
+    );
+  };
+
+  const rightIcon = props.rightIcon ? props.rightIcon : <></>;
+  return (
+    <div className={classes.root}>
+      <div className={classes.content}>
+        {props.numbers.map((num, index) =>
+          getLabeledNumber(index, num, props.labels[index], props.unit[index], props.primary, rightIcon),
```
julianrkung: Is there a unique non-index value we can eventually use (like an ID) we can use when we eventually pull from the airtable? If so, we can put it as a TODO to swap out using index as the key.

julianrkung: Since `props.label`, `props.unit`, and `props.numbers` are all connected via index, we could pass in an array of objects instead with properties {labels, unit, numbers} so that the relationship is clearer and easier to work with

src/screens/Customers/CustomerMain.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { useEffect, useState } from 'react';
+import { CustomerRecord } from '../../utils/airtable/interface';
+import { getAllCustomers } from '../../utils/airtable/requests';
+import { Link } from 'react-router-dom';
+
+interface CustomerMainProps {
+  match: any;
+}
+
+function CustomerMain(props: CustomerMainProps) {
```
julianrkung: Could we use `RouteComponentProps` here instead? https://stackoverflow.com/questions/48138111/what-typescript-type-should-i-use-to-reference-the-match-object-in-my-props . If not, it's no big deal, but I'd rather just refrain from using `any` if we're going to commit to typescript

src/components/OutlinedCardList.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+interface CardProps {
+  numbers: number[];
+  labels: string[];
+  unit: string[];
+  primary: boolean;
+  rightIcon?: JSX.Element;
+}
+
+export default function OutlinedColCard(props: CardProps) {
```
julianrkung: can we keep the name of component and the filename the component lives in in sync?

julianrkung: Also a bit of a NIT but if it's not a lot of trouble I think it's easier to keep abbreviations out of component names if possible. `OutlinedColumnCard` if `Col` stands for Column (but feel free to ignore if you feel strongly the other way)

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -0,0 +1,119 @@
...
+    },
+    buttonSecondary: {
+      borderRadius: '12px',
+      backgroundColor: theme.palette.secondary.main,
+    },
+  });
+
+interface CustomerProps {
+  classes: any;
+  location: any;
```
julianrkung, line 40: same comments as before about classes/match/location props

src/screens/Customers/CustomerRecords.tsx
```diff
@@ -0,0 +1,30 @@
...
+    <>
+      <BaseHeader leftIcon="backNav" />
+      <h3>Meter Readings</h3>
+      {meterReadings.map((meterReading: MeterReadingRecord, index) => (
+        <p key={index}>{`${meterReading.date} ${meterReading.reading}`}</p>
+      ))}
+
+      <h3>Invoices</h3>
+      {invoices.map((invoice: InvoiceRecord, index) => (
+        <p key={index}>{`${invoice.date} ${invoice.amount}`}</p>
```
julianrkung: are the rIds for meterReadings / invoices unique? If so, can we use them for keys instead of indices?

src/components/OutlinedCardList.tsx
```diff
@@ -0,0 +1,74 @@
...
+          <Typography variant="h3" color={primary ? 'primary' : 'inherit'}>
+            {number} {unit}
+          </Typography>
+        </div>
+        {rightIcon}
+      </div>
+    );
+  };
+
+  const rightIcon = props.rightIcon ? props.rightIcon : <></>;
```
julianrkung: does the styling break if we do null here instead of `<></>`? If not, returning null might be clearer than an empty element (if people aren't familiar with this in jsx)

julianrkung (changes requested):
Thank u for doing this u r truly best PL!! :^)

A couple other things I found:
If we go directly to the customer page, the bottom "Person" icon isn't lit up like it's supposed to be.
image

If I scroll down on a customer-specific page, I end up with the following:
image

Those 2 issues might be more navbar-related though, so we might want a separate PR for them.

src/screens/Customers/CustomerProfile.tsx
jenmhoang: - resolved for classes and match
- struggling to correctly type location, but since we're probably going to eventually replace this w redux store anyways, don't think it's worth digging deeper (unless you think it's important)

src/components/OutlinedCardList.tsx
jenmhoang: This is a base component that isn't connected to a particular record w an id but maps calculated numbers for a specific customer

src/screens/Customers/CustomerProfile.tsx
julianrkung: Not very important; thanks for making the initial change though!

julianrkung (approved):
Thanks Jen! :^)

PR #13, opened 2020-11-06
Add state related to online/offline and last time data pulled

julianrkung:
Redux now stores the last time it successfully pulled data.
To do this, it pings airlock every 10 seconds, and if a ping fails, it determines that the app is no longer online and stores that state in redux. (displayed on the home screen)

Frontend looks kinda jank, but wasn't in scope of this PR to do frontend.

Screenshots:
image

image


jenmhoang (approved):
lgtmmm, UserFields will prob be updated slightly w new schema but that can be fixed in the future

PR #14, opened 2020-11-13
Financial Summary Screens

mdterrenal:
## 🖋 PR Description

This PR adds in the financial summary screens as screenshotted below, and also pulls in data from Airtable to populate the fields. Currently a specific financial report is being pulled instead of the one associated with a site since I'm not quite sure how to go about that yet (does it need some sort of context so that we know which site we're pulling from? etc)

### 🤳Screenshots (if applicable)

screen1
screen2

### 🐞 Bug Report
warnings
Currently running into these warnings right now and I don't know enough about refs to understand what it's talking about :') But despite the warnings everything shows up properly. Also I'm not quite sure how your profit and Mee Panyar's profit is calculated based off of the total profit, so I just input manual zeroes in for now but that can be changed once that's figured out.

### 🧪 Testing

How do we test your feature?
I currently redirected the Incidents page to this screen since I don't think the link from the home page to this is available yet? This will have to be changed later on, but you can go to http://localhost:3000/financialsummary after running npm start to check these files out! Also I can't get Helvetica Neue on my computer so I tested with Arial so if someone would be able to test with the right font for me that would be wonderful :')

cc: @jenmhoang


src/screens/FinancialSummary/ConfirmModalContents.tsx
```diff
@@ -0,0 +1,106 @@
...
+  return (
+    <div className={classes.modalContents}>
+      <div className={classes.valueWrappers}>
+        <TextWrapper
+          labels={profitLabels.slice(0, 1)}
+          numbers={props.profitNumbers}
+          units={currencyUnits.slice(0, 1)}
+          bold
+        />
+        <Divider className={classes.invisibleDivider} />
```
julianrkung: is there a reason you used an invisible divider here instead of margin-ing the bottom?

src/screens/FinancialSummary/ConfirmModalContents.tsx
```diff
@@ -0,0 +1,106 @@
...
+        <TextWrapper
+          labels={profitLabels.slice(1)}
+          numbers={zeroes}
+          units={currencyUnits.slice(1)}
+          color="textSecondary"
```
julianrkung: If everything in TextWrapper is based on indices of these lists, do you think it would be cleaner to create objects with {unit, label, value...} key value pairs and mapping over that?

src/components/TextWrapper.tsx
```diff
@@ -0,0 +1,74 @@
...
+  color?: 'inherit' | 'primary' | 'textSecondary';
+  bold?: boolean;
+}
+
+export default function TextWrapper(props: TextWrapperProps): JSX.Element {
+  const classes = useStyles();
+
+  const getLabeledNumber = (number: number, label: string, index: number) => {
+    return (
+      <div key={index} className={classes.items}>
```
julianrkung: It's better if indices aren't used as keys. See https://reactjs.org/docs/lists-and-keys.html#keys and https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318 . Maybe we could use label instead if that's unique?

src/components/TextWrapper.tsx
```diff
@@ -0,0 +1,74 @@
...
+          </>
+        ) : (
+          <>
+            <Typography color={props.color ? props.color : 'inherit'} variant="h4">
+              {label}
+            </Typography>
+            <Typography color={props.color ? props.color : 'inherit'} variant="h4">
+              {number + (props.units ? props.units[index] : '')}
+            </Typography>
+          </>
```
julianrkung: I don't think we need to wrap such a large ternary for this. I believe something along the lines of

```
const typographyClasses = props.bold ? classes.bold : null
...
...
<Typography className={typographyClasses}...> ... < />
...
```

can be done

src/screens/FinancialSummary/FinancialSummary.tsx
```diff
@@ -0,0 +1,183 @@
...
+              title={'Financial Summary'}
+              labels={financialSummaryLabels.slice(0, 2)}
+              numbers={financialSummaryNumbers.slice(0, 2)}
+              units={financialSummaryUnits.slice(0, 2)}
+            />
+            <Divider className={classes.invisibleDivider} />
+            <TextWrapper
+              labels={financialSummaryLabels.slice(2)}
+              numbers={financialSummaryNumbers.slice(2)}
+              units={financialSummaryUnits.slice(2)}
```
julianrkung: Similar comment about objects potentially being cleaner

julianrkung (changes requested):
Sorry for getting to this late! Suggested some changes and found some bugs. Really good work :) 

This was also a pretty quick go-around, so I'll take a deeper look in a bit and hopefully pin down the source of the warnings

## Getting [Object object] for Total Remaining Owed
image

## Submit report / Confirm button disappears on click
image
image

julianrkung:
Looks like the source of the errors might be because we're using the wrong MaterialUI component. I found this in the Modal API docs https://material-ui.com/api/modal/ . It looks like it requires a children prop. We're probably better off using Dialog component.

image

src/components/TextWrapper.tsx
mdterrenal: ooh yeah definitely i made some changes to clean this up and will push once i address all the comments.

src/screens/FinancialSummary/ConfirmModalContents.tsx
mdterrenal: honestly was lazy and didn't want to add an extra style in, but ill remove the dividers and switch to a class on the textwrapper.

mdterrenal:
i couldn't recreate the issue with the buttons disappearing; they still remain on my end when i click on them. i will also work on addressing the object object issue once i look into it a bit more. i will also look into switching the modal into a dialog when i get the chance.

src/screens/FinancialSummary/ConfirmModalContents.tsx
mdterrenal: oh also another reason, the textwrapper would need to have a prop defined to pass in the class name and i didnt really know how to go about doing that, so stuck with this method instead.

mdterrenal:
issue with object object was a problem with airtable messing up with a calculation, fixed that and got a number out so hopefully that object object stops popping up

src/screens/FinancialSummary/ConfirmModalContents.tsx
julianrkung: Ok! Either way works just wanted to know if other possibilities were explored

src/styles/ThemeStyles.ts
```diff
@@ -36,6 +39,10 @@ export const theme = createMuiTheme({
       fontSize: '12px',
       fontWeight: 500,
     },
+    body1: {
```
jenmhoang: I think main already has a `body1` (changed h4 to body1) with a fontSize of 12 -- might need to change this to `body2` when merging

jenmhoang (approved):
Probably the biggest thing is just rethinking how objects are passed (echo-ing Julian's previous thoughts)! The rest looks good though, so for sake of getting styling in and unblocking other parts of the Financial Report flow, we can merge this in for now and rethink that in a future PR!

julianrkung (approved):
Good job! We can address any other concerns as they come up!

PR #15, opened 2020-11-14
Update Airtable Abstractions

jenmhoang:


src/screens/Customers/CustomerMain.tsx
```diff
@@ -18,7 +18,7 @@ function CustomerMain(props: RouteComponentProps) {
   return (
     <>
       {customers.map((customer, index) => (
-        <Link key={index} to={{ pathname: `${props.match.url}/${customer.rid}`, state: { customer: customer } }}>
+        <Link key={index} to={{ pathname: `${props.match.url}/${customer.name}`, state: { customer: customer } }}>
```
julianrkung, line 21: Fine for now. We might want to put a TODO to change this in the future since we aren't enforcing unique customer names

src/utils/airtable/interface.ts
```diff
@@ -1,183 +1,96 @@
...
 
 export type Row = {
   get: (key: string) => TableValues;
   getId: () => string;
 };
 
-interface Record {
+export interface UserRecord {
   rid: string;
-}
```
julianrkung, line 25: I don't think it matters since we aren't typing anymore, but is there a reason you removed the base record class? I liked the `XRecord extends Record` setup that was going on (but again, not important if we aren't using it)

src/utils/airtable/interface.ts
```diff
@@ -1,183 +1,96 @@
...
-export interface InventoryUpdateRecord extends Record {
-  quantity: number;
-  receiptPhoto: Airtable.Attachment[];
-  inventoryItemId: string[];
-  adminApproved: boolean;
-  siteId: string[];
-  amountPaid: number;
-  notes: string;
-  dateRecorded: string;
+  meterNumber: number;
```
julianrkung, line 48: Need rid also if we don't extend from base record

src/utils/airtable/interface.ts
```diff
@@ -1,183 +1,96 @@
...
-  incidentUpdateIds: string[];
-  incidentUpdates: IncidentUpdateRecord[];
-  isResolved: boolean;
-  resolutionDate: string;
-  resolutionDescription: string;
-  resolutionPhoto: Airtable.Attachment[];
-  dateRecorded: string;
-  siteId: string[];
+export interface CustomerUpdateRecord {
+  dateUpdated: string;
```
julianrkung, line 63: rid like CustomerRecord comment

src/utils/airtable/interface.ts
```diff
@@ -1,183 +1,96 @@
...
-  category: string[];
 }
 
-export interface IncidentUpdateRecord extends Record {
-  description: string;
-  photos?: Airtable.Attachment[];
-  dateRecorded: string;
-  incidentId: string[];
+export interface MeterReadingRecord {
+  reading: number;
```
julianrkung, line 70: rid like MeterReadingRecord

julianrkung (approved):
Thank you :^))) . Approved w/small suggestions. Probably want to wait on Maria's lgtm though since she is more familiar with the code

mdterrenal (approved):
everything looks good! also want to +1 on julian's comments above about the rid; i believe the transformRecord function within schema.ts sets the record's rid to the id from the pulled object from airtable (line 128 on github), but even without the rid defined as an attribute in the record type in interface.ts, there doesn't seem to be any issues after pulling data. im not sure if we need to have all the rids defined or not tho :')

src/utils/airtable/interface.ts
jenmhoang: talked via voice chat ab this -- removing rid template and only preserving for Users, Sites, and Tariffs for offline edit/add support

cc: @mdterrenal

PR #17, opened 2020-11-15
Sign in Authentication

julianrkung:
Added sign in authentication. A user should not be able to visit any AuthenticatedRoute unless they are logged in. They will be redirected to the `/login` route until they log in. The `/` route now redirects to login page if they are not logged in, and to Home page if they are logged in.

I couldn't find a sign out workflow on Figma, so to "sign out", one can just clear their local storage. This can be done via going to `devtools --> application --> "Clear Storage" --> "Clear site data" `


jenmhoang (approved):
king

PR #19, opened 2020-11-28
Meter Readings, Payments, and Customer Creation offline

julianrkung:
This PR allows meter readings, payments, and customers to be able to be created offline.

Workbox already allows Meter Readings / Payments to be created for customers that are already in the airtable. The main workflow that this PR addresses is adding meter readings / payments for a customer that isn't in the airtable yet (for example, the customer could've been created while offline and the user hasn't come back online yet).

The `customer` and `meterReading` data should look like the following:
```
Meter reading data looks like:     
const meterReading = {
      reading: 10,
      meterNumber: 10,
      date: moment().format('YYYY-MM-DD'),
      amountBilled: 100,
      billedById: getUserId(),
    };

Customer data looks like:     
const customerData = {
      name: customerName,
      meterNumber: meterNumber,
      tariffPlansId: tariffPlanId,
      sitesId: siteId,
      meterReadings: [],
      payments: [],
    };
```


jenmhoang (approved):
you never fail

PR #20, opened 2020-11-29
Tiffany/submitted records

cnnmon:
🖋 **PR Description**

Pulled from main, constructed financial records screen and components (FinancialInfo, PaymentCard).

🤳 **Screenshots**

image
image

🐞 **Bug Report**
Not that I know of right now!

🧪 **Testing**
http://localhost:3000/records


src/styles/ThemeStyles.ts
```diff
@@ -15,7 +15,7 @@ export const theme = createMuiTheme({
     divider: '#E5E5E5',
   },
   typography: {
-    fontFamily: 'Helvetica Neue',
+    fontFamily: 'Arial',
```
jenmhoang: Change this back from testing font so it's not merged into main!

src/screens/SubmittedRecords.tsx
```diff
@@ -0,0 +1,134 @@
...
+        <Typography className={classes.title} variant="h4">
+          Unpaid Reports
+        </Typography>
+        {dummyData.map((report: any, index) => (
+          <OutlinedColCard key={report.label} info={report} primary={false} rightIcon={getPaymentButtons()} />
+        ))}
+        <Typography className={classes.title} variant="h4">
+          Paid Reports
+        </Typography>
+        {dummyData.map((report: any, index) => (
```
jenmhoang: since you're not using index at all here, you can remove `index`! same applies to other mappings that don't actually use index

src/screens/SubmittedRecords.tsx
```diff
@@ -0,0 +1,134 @@
...
+    content: {
+      textAlign: 'left',
+      color: theme.palette.text.secondary,
+      padding: '0px 20px 80px',
+      width: '100%',
+      height: '80vh',
+      overflow: 'auto',
+    },
+    payButton: {
+      zIndex: -1,
```
jenmhoang: since this places the payment button underneath all other elements, it's now no longer clickable

if this was done because it was going over the NavBar, dwab, we can fix NavBar issues in a future PR

src/screens/SubmittedRecords.tsx
```diff
@@ -0,0 +1,134 @@
...
+          color="primary"
+        />
+        <SearchIcon className={classes.searchIcon} />
+      </div>
+    );
+  };
+
+  const getPaymentButtons = () => {
+    return (
+      <Button className={classes.payButton} color="primary" disableElevation={true}>Pay</Button>
```
jenmhoang: you might want to add `variant="contained"` to the buttons (others in addition to this one) to avoid that bug where it disappears upon clicking it

src/screens/SubmittedRecords.tsx
```diff
@@ -0,0 +1,134 @@
...
+    payButton: {
+      zIndex: -1,
+      borderRadius: '12px',
+      height: '30px',
+      width: '80px',
+      color: 'white',
+      backgroundColor: theme.palette.primary.main,
+    },
+    viewButton: {
+      zIndex: -1,
```
jenmhoang: same as above

src/screens/SubmittedRecords.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const getSearch = () => {
+    return (
+      <div className={classes.searchContainer}>
+        <TextField
+          className={classes.search}
+          size="small"
+          InputProps={{ style: { fontSize: 14 }, disableUnderline: true, autoComplete: 'off' }}
+          color="primary"
+        />
+        <SearchIcon className={classes.searchIcon} />
```
jenmhoang: you might be able to add icons in with the TextField component using `InputAdornments`, linked here: https://material-ui.com/components/text-fields/#input-adornments

src/screens/SubmittedRecords.tsx
```diff
@@ -0,0 +1,134 @@
...
+      <Button className={classes.viewButton} color="primary" disableElevation={true}>View</Button>
+    );
+  };
+
+  return (
+    <div className={classes.root}>
+      <BaseHeader leftIcon="backNav" title="Reports" />
+      {getSearch()}
+      <div className={classes.content}>
+        <Typography className={classes.title} variant="h4">
```
jenmhoang: I think this is meant to be `body1` instead of `h4`! right now the font-size is much larger than 12px

src/screens/SubmittedRecords.tsx
```diff
@@ -0,0 +1,134 @@
+import React from 'react';
+import BaseHeader from '../components/BaseComponents/BaseHeader';
+import { List, TextField, Typography, Button } from '@material-ui/core';
```
jenmhoang: Remove unused imports! (List)

src/components/FinancialInfo.tsx
```diff
@@ -0,0 +1,97 @@
...
+            </Typography>
+            <Typography className={classes.bolded} variant="h4">
+              Account Number
+            </Typography>
+            <Typography className={classes.bolded} variant="h4">
+              Account Name
+            </Typography>
+          </div>
+          <div className={classes.right}>
+            <Typography variant="h4">{props.bankName}</Typography>
```
jenmhoang: change all `h4` to `body1` as updated in ThemeStyles

src/components/FinancialInfo.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    root: {
+      color: theme.palette.text.primary,
+    },
+    title: {
+      color: theme.palette.text.secondary,
+    },
+    content: {
```
jenmhoang: this might need a `text-align = left`

src/components/FinancialInfo.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { makeStyles, createStyles, Theme } from '@material-ui/core/styles';
+import TextField from './TextField';
+
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    root: {
+      color: theme.palette.text.primary,
+    },
+    title: {
+      color: theme.palette.text.secondary,
```
jenmhoang: since this is being applied to a typography component, you can just remove this styling completely and instead write the component as `<Typography variant="h2" color="textSecondary"/>`

src/components/FinancialInfo.tsx
```diff
@@ -0,0 +1,97 @@
...
+          </div>
+          <div className={classes.right}>
+            <Typography variant="h4">{props.bankName}</Typography>
+            <Typography variant="h4">**** **** {props.accountNumber}</Typography>
+            <Typography variant="h4">{props.accountName}</Typography>
+          </div>
+        </div>
+        <Divider className={classes.divider} />
+        <div className={classes.itemWrapper}>
+          <div>
```
jenmhoang: this doesn't need to be wrapped in an extra div

src/components/PaymentCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+interface CardProps {
+  payment: number;
+  paid: number;
+}
+
+export default function FilledCard(props: CardProps) {
+  const classes = useStyles();
+  const getPaymentButtons = () => {
+    return (
+      <Button className={classes.button} color="primary" disableElevation={true}>
```
jenmhoang: `variant="contained"`

src/components/PaymentCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+      height: '20px',
+    },
+    button: {
+      borderRadius: '12px',
+      width: '120px',
+      color: 'white',
+      backgroundColor: theme.palette.primary.main,
+    },
+    paymentText: {
+      color: theme.palette.primary.main,
```
jenmhoang: this styling can be removed and can be assigned with `color="primary"` in the component

jenmhoang (dismissed):
some small changes but looks goood

src/components/FinancialInfo.tsx
```diff
@@ -0,0 +1,97 @@
...
+          </div>
+          <div className={classes.right}>
+            <Typography variant="h4" className={classes.balanceText}>
+              {props.balance} Ks
+            </Typography>
+          </div>
+        </div>
+      </div>
+      <div className={classes.fieldContainer}>
+        <TextField label="Payment Amount" bold={true} />
```
jenmhoang: This isn't taking in the right props

src/App.tsx
```diff
@@ -61,8 +64,18 @@ function App(isSignedIn: AppProps) {
...
+              <AuthenticatedRoute path="/customers" component={CustomerMain} exact />
+              <AuthenticatedRoute path="/inventory" component={Inventory} />
+              <AuthenticatedRoute path="/maintenance" component={Maintenance} />
+              <AuthenticatedRoute path="/incidents" component={Incidents} />
+              <AuthenticatedRoute path="/records" component={SubmittedRecords} />
+              <AuthenticatedRoute path={'/customers/:rid'} component={CustomerProfile} exact />
+              <AuthenticatedRoute path={'/customers/:rid/records'} component={CustomerRecords} />
+              <AuthenticatedRoute path="/customers/add" component={AddCustomer} />
+              <AuthenticatedRoute path="/customers/:rid/addMeter" component={AddMeterReading} />
+              <AuthenticatedRoute path={'/customers/:rid/edit'} component={EditCustomer} />
```
julianrkung: Is there a reason this was changed (it was probably just a weird merge thing)? If not can it be changed back and just add the `/records` endpoint inside of Container?

src/App.tsx
```diff
@@ -61,8 +63,18 @@ function App(isSignedIn: AppProps) {
...
-              <Route component={Container} />
+              <Route path="/login" component={Login} />
+              <AuthenticatedRoute path="/home" component={Home} />
+              <AuthenticatedRoute path="/customers" component={CustomerMain} exact />
+              <AuthenticatedRoute path="/inventory" component={Inventory} />
+              <AuthenticatedRoute path="/maintenance" component={Maintenance} />
+              <AuthenticatedRoute path="/incidents" component={Incidents} />
+              <AuthenticatedRoute path="/records" component={SubmittedRecords} />
+              <AuthenticatedRoute path={'/customers/:rid'} component={CustomerProfile} exact />
+              <AuthenticatedRoute path={'/customers/:rid/records'} component={CustomerRecords} />
```
julianrkung: I don't think "records" is a provocative enough name (it's not clear "records of what?") . We might want to just do `/financial-summaries`? Ashley is doing a small redesign of the financial summary workflow so we can land on a more precise endpoint later but we should note this down to change!

package.json
```diff
@@ -64,7 +64,7 @@
     "clean-cra-sw": "rm -f build/precache-manifest.*.js && rm -f build/service-worker.js",
     "build": "react-scripts build && npm run build-sw && npm run clean-cra-sw",
     "generate-schema": "generate-airtable-schema",
-    "prettier-format": "prettier --config .prettierrc 'src/**/*.{ts,tsx,css}' --write",
+    "prettier-format": "prettier --config .prettierrc \"src/**/*.{ts,tsx,css}\" --write",
```
julianrkung, line 67: Is there a reason this was changed for this PR?

src/App.tsx
```diff
@@ -22,6 +22,8 @@ import { ConnectedRouter } from 'connected-react-router';
 import AuthenticatedRoute from './components/authentication/AuthenticatedRoute';
 import { Switch, Route, Redirect } from 'react-router-dom';
 import { RootState } from './lib/redux/store';
+import LabelBottomNavigation from './components/BaseComponents/BottomNav';
```
julianrkung: Deleted unused (and expired)

src/App.tsx
```diff
@@ -22,6 +22,8 @@ import { ConnectedRouter } from 'connected-react-router';
...
+import SubmittedRecords from './screens/SubmittedRecords';
```
julianrkung: We should change the name for this. Maybe just `FinancialSummaries` or something like that

src/screens/Customers/AddMeterReading.tsx
```diff
@@ -51,7 +51,7 @@ function AddMeterReading(props: AddMeterReadingProps) {
             <Typography variant="body1" style={{ marginBottom: 15 }}>
               00.00.0000
             </Typography>
-            <TextField label={'New Meter Reading (kWh)'} id={'meter-reading'} primary={true} />
+            <TextField label={'New Meter Reading (kWh)'} />
```
julianrkung, line 54: Is this an extra change from a previous PR? This PR should only deal with the financial summary stuff afaik. Is there a reason this was changed? (if not can it be changed back)

src/screens/Customers/EditCustomer.tsx
```diff
@@ -28,11 +28,11 @@ function EditCustomer(props: EditCustomerProps) {
...
     <BaseScreen title="Edit Customer" leftIcon="backNav">
       <form noValidate className={classes.content}>
-        <TextField label={'Name'} id={'name'} primary={true} />
+        <TextField label={'Name'} />
         <Checkbox label={'Select if customer is inactive'} />
         <Checkbox label={'Select if customer has meter'} />
-        <TextField label={'Tariff Plan'} id={'tarrif-plan'} primary={true} />
-        <TextField label={'Reason'} id={'reason'} primary={true} />
+        <TextField label={'Tariff Plan'} />
+        <TextField label={'Reason'} />
```
julianrkung, line 35: Is this an extra change from a previous PR? This PR should only deal with the financial summary stuff afaik. Is there a reason this was changed? (if not can it be changed back)

src/components/TextField.tsx
```diff
@@ -1,39 +1,38 @@
 import React from 'react';
-import { withStyles, createStyles, Theme } from '@material-ui/core/styles';
+import { makeStyles, createStyles, Theme } from '@material-ui/core/styles';
```
julianrkung: Was this file accidentally changed? The props you pass into `TextField` from `SubmittedRecords` is based on the old diff. If it was accidentally changed, would you mind changing it back?

src/screens/SubmittedRecords.tsx
```diff
@@ -0,0 +1,136 @@
+import React from 'react';
```
julianrkung: Lets think of a better name for this! Also, if this screen is used in `FinancialSummary` and nowhere else, can we put it in the `screens/FinancialSummary/` directory?

src/screens/SubmittedRecords.tsx
```diff
@@ -0,0 +1,136 @@
...
+import BaseHeader from '../components/BaseComponents/BaseHeader';
+import { TextField, Typography, Button, InputAdornment } from '@material-ui/core';
```
julianrkung: Remove unused import (`InputAdornment`)

src/screens/SubmittedRecords.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  const getPaymentButtons = () => {
+    return (
+      <Button className={classes.payButton} variant="contained" color="primary" disableElevation={true}>Pay</Button>
+    );
+  };
+
+  const getViewButtons = () => {
+    return (
+      <Button className={classes.viewButton} variant="contained" color="primary" disableElevation={true}>View</Button>
```
julianrkung: With JSX, if you want to pass `true` to a prop, you can just write the prop name. Instead of 

```
<Button ... disableElevation={true}... />
```

You can just do 

```
<Button ... disableElevation ... />
```

https://reactjs.org/docs/jsx-in-depth.html#props-default-to-true lets just standardize the codebase to without the `={true}`?

src/screens/SubmittedRecords.tsx
```diff
@@ -0,0 +1,136 @@
+import React from 'react';
+import BaseHeader from '../components/BaseComponents/BaseHeader';
+import { TextField, Typography, Button, InputAdornment } from '@material-ui/core';
+import OutlinedColCard from '../components/OutlinedCardList';
```
julianrkung: Is there a reason this is aliased (`OutlinedColCard` instead of `OutlinedCardList`).  I remember that this component used to be called something similar Just wanna make sure it's not an oversight.

src/components/FinancialInfo.tsx
```diff
@@ -0,0 +1,90 @@
...
+);
+
+interface CardProps {
+  bankName: string;
+  accountNumber: number;
+  accountName: string;
+  balance: number;
+}
+
+export default function FilledCard(props: CardProps) {
```
julianrkung: Usually react components are named after their file name (or vice versa). Lets think of a better name for this component (and the props usually share the same name as the component appended with `Props`)

src/components/PaymentCard.tsx
```diff
@@ -0,0 +1,77 @@
...
+    },
+  }),
+);
+
+interface CardProps {
+  payment: number;
+  paid: number;
+}
+
+export default function FilledCard(props: CardProps) {
```
julianrkung: Same thing as FinancialInfo, we should look into better names

julianrkung (changes requested):
Is there a screen that displays FinancialInfo and PaymentCard components? I saw on the screenshot but I couldn't navigate to anything like that in the app

package.json
cnnmon: I use Windows so I need this for Git to work!

src/screens/SubmittedRecords.tsx
cnnmon: Honestly not sure (I think I was referencing that from previous files) so I changed it to be named OutlinedCardList too.

src/lib/airtable/schema.js
```diff
@@ -30,6 +30,7 @@ export const Columns = {
 		name: {name:`Name`, type:`text`},
 		userIds: {name:`Users`, type:`foreignKey-many`},
 		customers: {name:`Customers`, type:`custom-object`},
+		financialSummaries: {name: `FinancialSummaries`, type:`custom-object`},
```
julianrkung, line 33: This file is going to get regenerated every time we run the schema generator, and this file is going to get lost. 
I'll show you how to make a patch for it later so that we don't have to always manually re-enter this field whenever we regenerate

src/screens/FinancialSummary/FinancialSummaries.tsx
```diff
@@ -0,0 +1,151 @@
...
+        {paidFinancialSummaries.map((report: any, index: number) => (
+          <OutlinedCardList key={index} info={[report]} primary={false} rightIcon={getViewButtons()} />
+        ))}
+      </div>
+    </div>
+  );
+}
+
+const mapStateToProps = (state: RootState) => ({
+  financialSummaries: state.siteData.currentSite.financialSummaries || '',
```
julianrkung: Let's protect against the case that `currentSite` is null. If that's the case, we want `financialSummaries` to be an empty array, not an empty string

src/screens/FinancialSummary/FinancialSummaries.tsx
```diff
@@ -0,0 +1,151 @@
...
+    root: string;
+    content: string;
+    payButton: string;
+    viewButton: string;
+    searchIcon: string;
+    searchContainer: string;
+    search: string;
+    title: string;
+  };
+  financialSummaries?: any;
```
julianrkung: These should have type `FinancialSummaryRecord[]` and shouldn't be optional imo. We can deal with the optional case in our `mapStateToProps` function and pass an empty array in that case.

src/screens/FinancialSummary/FinancialSummaries.tsx
```diff
@@ -0,0 +1,151 @@
...
+  location: any;
+}
+
+function FinancialSummaries(props: FinancialSummariesProps) {
+  const { classes, financialSummaries } = props;
+  let paidFinancialSummaries: any[] = [];
+  let unpaidFinancialSummaries: any[] = [];
```
julianrkung: We should try our best to avoid `any` types especially since we're going to pass this codebase to developers so we want our code to be as readable as possible.

These lists have a specified format that they MUST adhere to since we pass them to `OutlinedCardList`, so let's make sure our types reflect that. The payload format is found [here](https://github.com/calblueprint/meepanyar/blob/main/src/components/OutlinedCardList.tsx#L32). We can create another interface for what `info` should look like and export it so that we can use it here. Should look something like this

```
// In OutlinedCardList.tsx
export interface CardPropsInfo {
  number: number;
  label: string;
  unit: string;
}

interface CardProps {
  info: CardPropsInfo[];
  ...
}

// In FinancialSummaries.tsx
(import the CardPropsInfo interface)
...
let paidFinancialSummaries: CardPropsInfo[] = [];
let unpaidFinancialSummaries: CardPropsInfo[] = [];
...
const entries : CardPropsInfo = {...};

```

Let me know if that makes sense!

src/screens/FinancialSummary/FinancialSummaries.tsx
```diff
@@ -0,0 +1,151 @@
...
+  };
+
+  const getViewButtons = () => {
+    return (
+      <Button className={classes.viewButton} variant="contained" color="primary" disableElevation={true}>View</Button>
+    );
+  };
+
+  return (
+    <div className={classes.root}>
```
julianrkung: Can we use the `BaseScreen` and `BaseScrollView` components instead of some of these custom divs? I think the components were made by Jen after this PR started, but it should abstract away the need to use `BaseHeader` and hopefully cut down on some custom styles. You can look at `CustomerMain` to see a decent example of how @kyleqhua used it, and look at what props it receives also. 

I think this should be a small-ish change, but let me know if it gets too out of hand and we can figure something else out.

src/screens/FinancialSummary/FinancialSummaries.tsx
```diff
@@ -0,0 +1,151 @@
...
+          Unpaid Reports
+        </Typography>
+        {unpaidFinancialSummaries.map((report: any, index: number) => (
+          <OutlinedCardList key={index} info={[report]} primary={false} rightIcon={getPaymentButtons()} />
+        ))}
+        <Typography className={classes.title} variant="body1">
+          Paid Reports
+        </Typography>
+        {paidFinancialSummaries.map((report: any, index: number) => (
+          <OutlinedCardList key={index} info={[report]} primary={false} rightIcon={getViewButtons()} />
```
julianrkung: At first I was a little confused about why we're giving `OutlinedCardList` a list of single reports (since we could just be passing all the `paidFinancialSummaries` into the `info` prop). I understand now that it's for style purposes (to make each report its own card), but we might want to clarify that via a comment (on both the unpaid and paid summaries) because otherwise it just looks like weirdly written code HAHA

src/screens/FinancialSummary/FinancialSummaries.tsx
```diff
@@ -0,0 +1,151 @@
...
+    );
+  };
+
+  const getPaymentButtons = () => {
+    return (
+      <Button className={classes.payButton} variant="contained" color="primary" disableElevation>Pay</Button>
+    );
+  };
+
+  const getViewButtons = () => {
```
julianrkung: Both of these functions should be singular since it returns only 1 button

src/screens/FinancialSummary/FinancialSummaries.tsx
```diff
@@ -0,0 +1,151 @@
...
+      });
+  };
+
+  formatFinancialSummaries(financialSummaries);
+
+  const getSearch = () => {
+    return (
+      <div className={classes.searchContainer}>
+        <TextField
+          className={classes.search}
```
julianrkung: The Figma has a date range rather than a text search. 

We don't have to have functionality for the date range now, but if it's not too much trouble could the frontend portion of it be added (no functionality required yet)? If you don't have time, we can delete this search part for now and just add in a date range later in a future sprint.

image

julianrkung (commented):
Some small changes (mostly NIT-like code health and typing stuff) and I think we'll be good after this iteration :)

GOOD STUFF!

src/screens/FinancialSummary/FinancialSummaries.tsx
```diff
@@ -0,0 +1,151 @@
+import React from 'react';
```
julianrkung: Sorry 1 more thing. To keep the naming convention consistent with `Customers` (which has `CustomerMain`), can we rename this `FinancialSummariesMain`? I think I goofed on the file name when I told you to rename to `FinancialSummaries` instead of `FinancialSummariesMain`, sorry about the extra work!

src/screens/FinancialSummary/FinancialSummaries.tsx
```diff
@@ -0,0 +1,151 @@
...
+}
+
+function FinancialSummaries(props: FinancialSummariesProps) {
+  const { classes, financialSummaries } = props;
+  let paidFinancialSummaries: any[] = [];
+  let unpaidFinancialSummaries: any[] = [];
+
+  const formatFinancialSummaries = (summaries: FinancialSummaryRecord[]) => {
+      summaries.forEach(summary => {
+        const entry = { number: summary.totalAmountBilled, label: summary.lastUpdated, unit: 'MMK'};
```
julianrkung: I think according to the Figma we want the large text to show period rather than totalAmountBilled. For the period, let's just use the month and year? I think there's a field in financial summaries that already has the period as a string, we might be able to just show that

image

src/screens/FinancialSummary/FinancialSummaries.tsx
cnnmon: I'll have to look into this more, for now I'll delete the search. Are there any Material UI samples out there for what we're looking for?

julianrkung (approved):
Thanks for iterating so quickly, LGTM! :)

PR #21, opened 2020-12-05
Tiffany/customer data with redux

cnnmon:
## 🖋 **PR Description**

This PR grabs customer data from redux and uses them throughout CustomerMain, CustomerProfile, and CustomerRecords pages according to currentSite.

🤳 **Screenshots**

🐞 **Bug Report**
Not that I know of right now!

🧪 **Testing**
http://localhost:3000/customers


src/components/FinancialInfo.tsx
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
```
jenmhoang: this file looks like it was leftover from your financial records frontend PR -- remove it from this branch/PR to keep them separate!

src/screens/SubmittedRecords.tsx
```diff
@@ -0,0 +1,148 @@
+import React from 'react';
```
jenmhoang: same message as before ab leftover files from separate PR

src/screens/Customers/CustomerMain.tsx
```diff
@@ -2,28 +2,31 @@ import React, { useEffect, useState } from 'react';
 import { CustomerRecord } from '../../utils/airtable/interface';
 import { getAllCustomers } from '../../utils/airtable/requests';
```
jenmhoang: remove unused imports

src/screens/Customers/CustomerMain.tsx
```diff
@@ -2,28 +2,31 @@ import React, { useEffect, useState } from 'react';
...
-  return (
-    <>
-      {customers.map((customer, index) => (
-        <Link key={index} to={{ pathname: `${props.match.url}/${customer.name}`, state: { customer: customer } }}>
-          <h3>{customer.name}</h3>
-        </Link>
-      ))}
-    </>
-  );
+class CustomerMain extends React.Component<CustomerProps, {}> {
```
jenmhoang: Is there a reason why you changed this from a functional component to a class component?

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -36,92 +37,137 @@ const styles = (theme: Theme) =>
...
 }
 
-function CustomerProfile(props: CustomerProps) {
-  const { classes, match } = props;
-  const customer: CustomerRecord = props.location.state.customer;
+class CustomerProfile extends React.Component<CustomerProps, {}> {
+  render() {
+    const { classes, match } = this.props;
+    const customer = this.props.location.state.customer;
+    const meterReadings = customer.meterReadings;
```
jenmhoang: `meterReadings` doesn't seem to be used anywhere

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -36,92 +37,137 @@ const styles = (theme: Theme) =>
...
+      return period == lastPer;
+    };
+
+    const getStartingMeter = (customer: CustomerRecord): MeterReadingRecord | undefined => {
+      return customer.meterReadings.find(function (mr: MeterReadingRecord) {
+        return isStartingMeter(mr);
+      });
+    };
+
+    const getAmountBilled = (customer: CustomerRecord, currReading: number, startingMeter = 0) => {
```
jenmhoang: remove unused functions

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -36,92 +37,137 @@ const styles = (theme: Theme) =>
...
+      const tariffPlan = customer.tariffPlans[0];
+      const periodUsage = currReading - startingMeter;
+      return tariffPlan.fixedTariff + tariffPlan.tariffByUnit * periodUsage;
+    };
+
+    // data retrieval
+    const currReading: MeterReadingRecord | undefined = getCurrentReading(customer);
+    const startingMeter: MeterReadingRecord | undefined = getStartingMeter(customer);
+    const currReadingAmount = currReading? currReading.amountBilled : 0;
```
jenmhoang: default might want to be changed from '0' to '-' based on the discussion over design we had during hackathon

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -36,92 +37,137 @@ const styles = (theme: Theme) =>
...
+
+    const getStartingMeter = (customer: CustomerRecord): MeterReadingRecord | undefined => {
+      return customer.meterReadings.find(function (mr: MeterReadingRecord) {
+        return isStartingMeter(mr);
+      });
+    };
+
+    const getAmountBilled = (customer: CustomerRecord, currReading: number, startingMeter = 0) => {
+      const tariffPlan = customer.tariffPlans[0];
+      const periodUsage = currReading - startingMeter;
```
jenmhoang: You should probably add a check on whether or not the current reading is valid or not (if there is no current reading, there should also be no period usage -- this also prevents it from displaying negative numbers when no reading exists)

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -36,92 +37,137 @@ const styles = (theme: Theme) =>
...
+  match: { params: string; isExact: boolean; path: string; url: string; }
   classes: { root: string; content: string; header: string; button: string; buttonSecondary: string };
+  siteName: string;
   location: any;
 }
 
-function CustomerProfile(props: CustomerProps) {
-  const { classes, match } = props;
-  const customer: CustomerRecord = props.location.state.customer;
+class CustomerProfile extends React.Component<CustomerProps, {}> {
```
jenmhoang: same as before ab functional components

src/screens/Customers/CustomerRecords.tsx
```diff
@@ -23,20 +25,22 @@ interface CustomerRecordsProps extends RouteComponentProps {
   location: any;
 }
 
-function CustomerRecords(props: CustomerRecordsProps) {
-  const { classes } = props;
-  const payments: PaymentRecord[] = props.location.state.payments;
-  const invoices: MeterReadingRecord[] = props.location.state.invoices;
+class CustomerRecords extends React.Component<CustomerRecordsProps, {}> {
```
jenmhoang: same as before ab functional components

src/screens/SubmittedRecords.tsx
jenmhoang: same as before about removing leftover files from previous PRs

jenmhoang (dismissed):

src/lib/redux/siteData.ts
```diff
@@ -67,4 +69,16 @@ const getPaymentsForCustomer = async (customer: CustomerRecord): Promise<Payment
...
+  const tariffIds = customer.tariffPlansId;
+  let tariffPlans: TariffPlanRecord[] = [];
+
+  if (tariffIds) {
+    tariffPlans = await getTariffPlansByIds(tariffIds);
+  }
+  
+  return tariffPlans;
+};
+
```
julianrkung: Data fetching like for tariff plans were moved to https://github.com/calblueprint/meepanyar-node/blob/main/resolvers/index.js

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,23 +1,18 @@
...
-    getCustomers();
-  }, []);
-
-  const [customers, setCustomers] = useState<CustomerRecord[]>([]);
-
-  const getCustomers = async () => {
-    const customers: CustomerRecord[] = await getAllCustomers();
-    setCustomers(customers);
-  };
+interface CustomerProps {
```
julianrkung: rename `CustomerMainProps` just to keep with ts convention

src/screens/Customers/CustomerMain.tsx
```diff
@@ -26,4 +21,9 @@ function CustomerMain(props: RouteComponentProps) {
   );
 }
 
-export default CustomerMain;
+const mapStateToProps = (state: RootState) => {
+  let customers = state.siteData.currentSite.customers;
```
julianrkung: just to ensure we're defensive w/code, let's do null checks to make sure we have a `currentSite` and whatnot (and if not customers is an empty list so the rest of the code doesnt break)

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -36,14 +37,63 @@ const styles = (theme: Theme) =>
...
   location: any;
 }
 
 function CustomerProfile(props: CustomerProps) {
   const { classes, match } = props;
-  const customer: CustomerRecord = props.location.state.customer;
+  const customer = props.location.state.customer;
 
+  // helper functions for getting data
+  const isCurrentReading = (mr: MeterReadingRecord): boolean => {
```
julianrkung: These looks like they'd be generally useful in the future outside of this component. If helper functions are potentially useful outside of this screen / component, create a utils file (like a `customerUtils` where you have functions that take in a `customerRecord` and give you certain things like their currentMeterReading, etc... a `meterReadingsUtil` file might also be helpful) for it and put all the functions in there and import them to use in here!

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -36,14 +37,63 @@ const styles = (theme: Theme) =>
...
+    const period: number = parseInt(mr.date.slice(5, 7));
+    const currPer: number = moment().month() + 1;
```
julianrkung: Lets move this period calculation and all moment-related stuff to `momentUtils` or another utils file as functions so that it's only written once since it's a little confusing and can easily be messed up (especially the slice by 5 & 7, adding 1 to current period, and the mod 12 in `isStartingMeter`). I think `getCurrentPeriod` and `getPreviousPeriod` would be really helpful helper functions that add a layer of abstraction over working with moment.

By abstracting away date calculations into functions in util functions that we just call, we reduce any type of errors that occur (off-by-1 or 0-index) since calculations will then always be done in the same way.

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -36,14 +37,63 @@ const styles = (theme: Theme) =>
...
+  const getStartingMeter = (customer: CustomerRecord): MeterReadingRecord | undefined => {
+    return customer.meterReadings.find(function (mr: MeterReadingRecord) {
+      return isStartingMeter(mr);
+    });
+  };
+
+  const getPeriodAndBilled = (customer: CustomerRecord, currReading: MeterReadingRecord, startingMeter: MeterReadingRecord | undefined) => {
+    const tariffPlan = customer.tariffPlans[0];
+    const periodUsage = currReading.amountBilled - (startingMeter? startingMeter.amountBilled : 0);
+    const amountBilled = tariffPlan.fixedTariff + tariffPlan.tariffByUnit * periodUsage;
```
julianrkung: Shouldn't periodUsage be based off of the `reading` field instead of amountBilled?

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -36,14 +37,63 @@ const styles = (theme: Theme) =>
...
+    return period == lastPer;
+  };
+
+  const getStartingMeter = (customer: CustomerRecord): MeterReadingRecord | undefined => {
+    return customer.meterReadings.find(function (mr: MeterReadingRecord) {
+      return isStartingMeter(mr);
+    });
+  };
+
+  const getPeriodAndBilled = (customer: CustomerRecord, currReading: MeterReadingRecord, startingMeter: MeterReadingRecord | undefined) => {
```
julianrkung: `getPeriodUsageAndAmountBilled` is a more clear name because `Period` standalone is confusing since it means something different. I would also decouple this into a `getPeriodUsage` and `getAmountBilled` since everything else is decoupled

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -118,10 +157,16 @@ function CustomerProfile(props: CustomerProps) {
...
+        <OutlinedColCard info={meterInfo} primary={false} />
       </div>
     </div>
   );
 }
 
-export default withStyles(styles)(CustomerProfile);
+const mapStateToProps = (state: RootState) => {
+  let customers = state.siteData.currentSite.customers;
+  let siteName = state.siteData.currentSite.name;
```
julianrkung: We should handle `currentSite` being potentially null just to be defensive (at the very least we shouldn't crash!). We can just set these to an empty list and empty string (or some placeholder like `Site Name`)

julianrkung (changes requested):
Really great stuff!! :) Would you mind updating the branch so I can run it locally and check everything out? Looks really good, just small changes

src/screens/Customers/CustomerMain.tsx
cnnmon: I saw that this file changed a lot in main, should I still make this change?

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -85,22 +104,11 @@ function CustomerProfile(props: CustomerProps) {
     );
   };
 
-  //dummy data
-  //TODO: pull from Airtable and hold information in the form of Records as defined in the schema
```
julianrkung, line 89: I can not express to you the happiness getting rid of this dummy data fills me with.

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -33,13 +36,29 @@ const styles = (theme: Theme) =>
...
   const customer: CustomerRecord = props.location.state.customer;
 
+  // data retrieval
+  const UNDEFINED_AMOUNT = '-';
+  const currReading: MeterReadingRecord | undefined = getCurrentReading(customer);
+  const startingMeter: MeterReadingRecord | undefined = getStartingMeter(customer);
+  const periodUsage = currReading ? getPeriodUsage(currReading, startingMeter) : UNDEFINED_AMOUNT;
+  const amountBilled = currReading ? getAmountBilled(currReading) : UNDEFINED_AMOUNT;
+
+  const meterInfo = [
```
julianrkung: The explicit type comment on your other PR applies to this as well. https://github.com/calblueprint/meepanyar/pull/20/files#r581504061 !

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -114,11 +122,15 @@ function CustomerProfile(props: CustomerProps) {
...
+          <OutlinedColCard info={meterInfo} primary={false} />
         </div>
       </BaseScrollView>
     </BaseScreen>
   );
 }
 
-export default withStyles(styles)(CustomerProfile);
+const mapStateToProps = (state: RootState) => ({
+  currentSite: state.siteData.currentSite || '',
```
julianrkung: This should default to an empty object instead, probably so we don't error when doing `currentSite.name` and whatnot.

src/screens/Customers/customerUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+
+export const getPeriodUsage = (currReading: MeterReadingRecord, startingMeter: MeterReadingRecord | undefined) => {
+  const periodUsage = currReading.reading - (startingMeter? startingMeter.reading : 0);
+  return periodUsage;
+}
+
+export const getAmountBilled = (currReading: MeterReadingRecord) => {
+  //const tariffPlan = customer.tariffPlans[0];
+  //const amountBilled = tariffPlan.fixedTariff + tariffPlan.tariffByUnit * getPeriodUsage(customer, currReading, startingMeter);
```
julianrkung: delete commented code if it's not being used

src/lib/moment/momentUtils.ts
```diff
@@ -5,3 +5,16 @@ import moment from 'moment';
...
+export const getThisPeriod = (date: string) => {
+  return parseInt(date.slice(5, 7));
+}
+
+export const getCurrentPeriod = () => {
+  return moment().month() + 1;
+}
+
+export const getLastPeriod = () => {
+  return moment().month();
```
julianrkung: Do these functions still work for December --> January and January --> December transitions? We should probably define a period as year and month so its unambiguous

src/lib/moment/momentUtils.ts
```diff
@@ -5,3 +5,16 @@ import moment from 'moment';
 export const formatUTCDateStringToLocal = (dateStringUTC: string): string => {
   return moment(dateStringUTC).local().format('YYYY-MM-DD HH:mm:ss');
 };
+
+// Returns month from date
+export const getThisPeriod = (date: string) => {
+  return parseInt(date.slice(5, 7));
+}
```
julianrkung: `getPeriodFromDate` is probably a better name! Also we might want to grab year here as well

src/screens/Customers/customerUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+//Returns last reading in the period
+export const getCurrentReading = (customer: CustomerRecord): MeterReadingRecord | undefined => {
+  return customer.meterReadings.find(function (mr: MeterReadingRecord) {
+    return isCurrentReading(mr);
+  });
+};
+
+const isStartingMeter = (mr: MeterReadingRecord): boolean => {
+  const period: number = getThisPeriod(mr.date) % 12;
+  const lastPer: number = getLastPeriod();
```
julianrkung: I'm curious about the case where a user forgets to do make a reading for last period. Should we use last last period? I'll write the NPO to ask, but leaving comment here to document

src/screens/Customers/customerUtils.ts
```diff
@@ -0,0 +1,39 @@
+import { CustomerRecord, MeterReadingRecord } from '../../lib/airtable/interface';
+import { getThisPeriod, getCurrentPeriod, getLastPeriod } from '../../lib/moment/momentUtils';
+
+const isCurrentReading = (mr: MeterReadingRecord): boolean => {
+    const period: number = getThisPeriod(mr.date);
```
julianrkung: Is there a reason that `period` is mod 12 on line 18 but not here? If a mod is required, let's stuff it inside of `getThisPeriod` so that it's abstracted away whenever someone calls it and people don't have to remember to mod

src/screens/Customers/customerUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const isStartingMeter = (mr: MeterReadingRecord): boolean => {
+  const period: number = getThisPeriod(mr.date) % 12;
+  const lastPer: number = getLastPeriod();
+  return period == lastPer;
+};
+
+//Returns
+export const getStartingMeter = (customer: CustomerRecord): MeterReadingRecord | undefined => {
+  return customer.meterReadings.find(function (mr: MeterReadingRecord) {
```
julianrkung: Does `customer.meterReadings.find(isStartingMeter)` work the same?

src/screens/Customers/customerUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const isCurrentReading = (mr: MeterReadingRecord): boolean => {
+    const period: number = getThisPeriod(mr.date);
+    const currPer: number = getCurrentPeriod();
+    return period == currPer;
+};
+
+//Returns last reading in the period
+export const getCurrentReading = (customer: CustomerRecord): MeterReadingRecord | undefined => {
+  return customer.meterReadings.find(function (mr: MeterReadingRecord) {
```
julianrkung: Does `customer.meterReadings.find(isCurrentReading)` the same?

src/screens/Customers/customerUtils.ts
```diff
@@ -0,0 +1,39 @@
+import { CustomerRecord, MeterReadingRecord } from '../../lib/airtable/interface';
+import { getThisPeriod, getCurrentPeriod, getLastPeriod } from '../../lib/moment/momentUtils';
+
+const isCurrentReading = (mr: MeterReadingRecord): boolean => {
+    const period: number = getThisPeriod(mr.date);
+    const currPer: number = getCurrentPeriod();
+    return period == currPer;
```
julianrkung: js comparisons are usually `===` not `==`

src/screens/Customers/customerUtils.ts
```diff
@@ -0,0 +1,39 @@
+import { CustomerRecord, MeterReadingRecord } from '../../lib/airtable/interface';
```
julianrkung: We want to be consistent with function naming. We should pick a name between `reading` and `meter` and stick with it. Right now we're using both (ex: `isCurrentReading` and `isStartingMeter`) which can be annoying / confusing

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -33,13 +36,29 @@ const styles = (theme: Theme) =>
...
 
 function CustomerProfile(props: CustomerProps) {
-  const { classes, match } = props;
+  const { classes, match, currentSite } = props;
   const customer: CustomerRecord = props.location.state.customer;
 
+  // data retrieval
+  const UNDEFINED_AMOUNT = '-';
+  const currReading: MeterReadingRecord | undefined = getCurrentReading(customer);
+  const startingMeter: MeterReadingRecord | undefined = getStartingMeter(customer);
```
julianrkung: We want to be consistent with variable naming also (I commented the same thing but about function names in `customerUtil`). For `reading` vs `meter` lets stick with one and be consistent so those reading won't be confused on if there's a difference between a meter and a reading

src/lib/airtable/schema.js
```diff
@@ -45,6 +46,7 @@ export const Columns = {
 	"Customers": {
 		name: {name:`Name`, type:`text`},
 		meterNumber: {name:`Meter Number`, type:`number`},
+		tariffPlans: {name: `TariffPlans`, type: `custom-object`},
```
julianrkung: delete if unnecessary

src/lib/airtable/schema.js
```diff
@@ -30,6 +30,7 @@ export const Columns = {
 		name: {name:`Name`, type:`text`},
 		userIds: {name:`Users`, type:`foreignKey-many`},
 		customers: {name:`Customers`, type:`custom-object`},
+		financialSummaries: {name: `FinancialSummaries`, type:`custom-object`},
```
julianrkung, line 33: We don't need this for this particular PR, right?

julianrkung (commented):
Let's talk with Ashley about starting meter, current period, etc so that everyone's on the same page because I'm a bit confused and can't do a thorough review without it.

In the meantime, I gave some code quality suggestions and thoughts about edge cases. I'll review during worksession after I get a better understanding!

src/screens/Customers/CustomerProfile.tsx
cnnmon: Making sure, would I do that like this?

```
  currentSite: state.siteData.currentSite || {
    rid: '',
    name: '',
    customerIds: [],
    customers: [],
    financialSummaryIds: [],
    financialSummaries: [],
  },
});
```

julianrkung: That's good! I think a better place to put the definition of the empty site object is as a const in `siteDataSlice.ts` (called like `EMPTY_SITE` or something like that). Then, we can export it and import it here to be used. That way, all empty sites we use are standardized and if the site field changes we only need to change it in 1 place to propagate it

src/lib/moment/momentUtils.ts
```diff
@@ -5,3 +5,24 @@ import moment from 'moment';
...
+  return year + "-" + month.toString().padStart(2, "0");
+}
+
+export const getLastPeriod = () => {
+  const year = moment().year();
+  let month = moment().month();
+  if (month === 0) {
+    month = 12;
+  }
+  return year + "-" + month.toString().padStart(2, "0");
```
julianrkung, line 29: If the current month is January we need the last period to be the 12th month of the previous year

src/lib/moment/momentUtils.ts
```diff
@@ -5,3 +5,24 @@ import moment from 'moment';
...
+
+export const getCurrentPeriod = () => {
+  const year = moment().year();
+  const month = moment().month() + 1;
+  return year + "-" + month.toString().padStart(2, "0");
+}
+
+export const getLastPeriod = () => {
+  const year = moment().year();
+  let month = moment().month();
```
julianrkung: It might be confusing that `moment().month()` is equal to `previousMonth` because of 0 index. We could add comments to make sure this comes across in the code

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -114,11 +122,22 @@ function CustomerProfile(props: CustomerProps) {
...
 
-export default withStyles(styles)(CustomerProfile);
+const mapStateToProps = (state: RootState) => ({
+  currentSite: state.siteData.currentSite || {
+    rid: '',
+    name: '',
+    customerIds: [],
+    customers: [],
+    financialSummaryIds: [],
+    financialSummaries: [],
```
julianrkung: https://github.com/calblueprint/meepanyar/pull/21#discussion_r582554429 commented on thread from last review

src/screens/Customers/customerUtils.ts
```diff
@@ -0,0 +1,34 @@
...
+//Returns last reading in the period
+export const getCurrentReading = (customer: CustomerRecord): MeterReadingRecord | undefined => {
+  return customer.meterReadings.find(isCurrentReading);
+};
+
+const isStartingReading = (mr: MeterReadingRecord): boolean => {
+  const period: string = getPeriodFromDate(mr.date);
+  const lastPer: string = getLastPeriod();
+  return period === lastPer;
+};
```
julianrkung: Doesn't this assume that the startingReading was from last period? I think we should have a broader definition of start reading: "The latest meter reading NOT from the current period". We should handle the case where a User is a little late to their readings or forgets a month (so there's no reading for the last period). 

This might require restructuring some of the functions

julianrkung (commented):
The code is really easy to read now! Small definition changes and then I'll check the frontend (the numbers will show up) and then we can merge

src/lib/utils/customerUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+// Returns last reading in the period
+export const getCurrentReading = (customer: CustomerRecord): MeterReadingRecord | undefined => {
+  return customer.meterReadings.find(isCurrentReading);
+};
+
+const isStartingReading = (mr: MeterReadingRecord): boolean => {
+  const period: string = getPeriodFromDate(mr.date);
+  const lastPer: string = getLastPeriod();
+  return period === lastPer;
+};
```
julianrkung: Let's delete this function if it's not being used so people don't get confused!

julianrkung (approved):
Approved! Feel free to merge after you delete the unused function!

Really quality work, code is PRISTINE :)

PR #22, opened 2020-12-06
Clean-up and Base Components

jenmhoang:
- Navbar no longer shows up on login screen
- added new `BaseScreen` component: content is confined to be between Header and Navigation (no overflow, scroll support)
- added `BaseScroll` component for content that should be in a scrollable container
- fixed nested scrolling in Records screen
- fixed Navbar bugs (width too large, didn't stay fixed to bottom)
- removed App.css file


src/index.tsx
```diff
@@ -3,7 +3,7 @@ import ReactDOM from 'react-dom';
 import { Provider } from 'react-redux';
 import { PersistGate } from 'redux-persist/integration/react';
 
-import './index.css';
+// import './index.css';
```
julianrkung: rm if not used

src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -57,22 +67,24 @@ function BaseHeader(props: HeaderProps) {
...
-        <Toolbar>
-          {props.leftIcon ? icons[props.leftIcon] : null}
-          {props.title ? (
-            <Typography className={classes.title} variant="h2">
-              {props.title}
-            </Typography>
-          ) : null}
-          {props.rightIcon ? icons[props.rightIcon] : <div className={classes.emptySpace} />}
-        </Toolbar>
-      </AppBar>
```
julianrkung, line 75: Remove the Toolbar and AppBar imports that aren't used anymore.

src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -57,22 +67,24 @@ function BaseHeader(props: HeaderProps) {
     menu: getIcon(history.goBack, <MenuIcon />), //replace history.goBack with correct functions
     backNav: getIcon(history.goBack, <ArrowBackIosIcon />),
     edit: getIcon(navigateToEdit, <CreateIcon />),
-    user: getIcon(history.goBack, <AccountCircleIcon />),
+    user: getIcon(history.goBack, <AccountCircleIcon className={classes.grey} />),
```
julianrkung, line 69: It's not really a big deal but imo it'd be cleaner to expose these elements themselves so that a developer can choose the elements to pass down rather than a string placeholder that ends up mapping to that element

ex: `<BaseScreen leftIcon={Icons.BACK_NAV} title={customer.name} rightIcon={Icons.EDIT} match={match}>` instead of `<BaseScreen leftIcon="backNav" title={customer.name} rightIcon="edit" match={match}>`.

Not a big deal though right now so feel free to ignore or put as TODO if you think it should be done that way.

julianrkung (approved):
You are the best. Left some nits that we can address later if you want

PR #12, opened 2020-11-02
Kyle/home

kyleqhua:
🖋 PR Description
Home page. Can select a site and see changes. Need to fix styling to search dropdown and add wifi icon.

🤳Screenshots (if applicable)

ezgif-4-0a6cf9b0f464

🐞 Bug Report

🧪 Testing

PR #24, opened 2020-12-10
Updated home page styling

jenmhoang:
- branched off of Kyle's initial home page styling (#12)
- updated home page styling to latest figma design
- functional menu/dropdown for selecting sites
- menu buttons navigate to appropriate pages
- show offline/online (changing icons)
localhost_3000_home(Galaxy S5)
localhost_3000_home(Galaxy S5) (1)
localhost_3000_home(Galaxy S5) (2)


src/screens/Home/components/SiteMenu.tsx
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  const closeMenu = () => {
+    setAnchorEl(null);
+  };
+
+  return (
+    <div className={classes.root}>
+      <Typography variant="h1">
+        {currentSite.name}
```
julianrkung: Heads up, this part errors if you clear the redux store (application cache) and do a fresh login. The fix should be done in another PR (I can do it). Basically, we try to use `currentSite` before it's set in Redux (since the HTTP call takes a bit of time). For now, we can just make the code more defensive by doing `currentSite && currentSite.name`. If you could put a `TODO` with a comment, I can prioritize addressing it by adding a "loading" screen tomorrow.

src/screens/Home/components/SiteMenu.tsx
```diff
@@ -0,0 +1,83 @@
...
+        <IconButton className={classes.icon} onClick={openMenu}>
+          <ArrowDropDownIcon color="primary" />
+        </IconButton>
+        <Menu anchorEl={anchorEl} keepMounted open={Boolean(anchorEl)} onClose={closeMenu}>
```
julianrkung, line 63: The following warning pops up if I open the little menu thing. I don't know the fix immediately. I'll let you make a judgement call on if this should be fixed in this PR or just added to list of stuff to address next sem.

image

src/screens/Home/Home.tsx
```diff
@@ -0,0 +1,87 @@
...
+}
+
+const mapStateToProps = (state: RootState) => {
+  let lastUpdated = '';
+
+  if (state.userData.lastUpdated) {
+    lastUpdated = formatUTCDateStringToLocal(state.userData.lastUpdated);
+  }
+  const isOnline = state.userData.isOnline;
+  const currentSite = state.siteData.currentSite;
```
julianrkung, line 82: IMO it would be cleaner to grab the `currentSite` in `SiteMenu` rather than passing it here since it's already in the redux store and it's not being used anywhere else in this component. We also grab `sites` in the `SiteMenu` so it'd be better to be consistent about how we grab i think.

src/screens/Home/components/HomeMenuItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+  amount: number;
+};
+
+interface HomeMenuItemProps {
+  label: string;
+  amount?: number;
+  sublabels?: sublabel[];
+  noBadge?: boolean;
+  classes: { root: string; content: string; noAlert: string; sublabel: string; icon: string; grey: string };
+}
```
julianrkung, line 18: Naming aside, this entire component is really clean and easy to read. It sparks joy. :) thank u.

src/screens/Home/components/HomeMenuItem.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { withStyles, createStyles, Theme } from '@material-ui/core/styles';
+
+type sublabel = {
+  label: string;
+  amount: number;
+};
+
+interface HomeMenuItemProps {
+  label: string;
+  amount?: number;
```
julianrkung, line 14: Maybe `amount` isn't the best variable name here / for sublabel (or maybe the component itself is named weird? I think it's odd that a component named `HomeMenuItem` takes in an `amount` prop to begin with. I can't think of a better off the top of my head but I will keep thinking. If you can't think of a better name either we could also just put this on the backburner as well (or just deal with the weird naming)

src/screens/Home/Home.tsx
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
```
julianrkung, line 1: u r such a good developer please don't leave... :( you're SOO GOOD!!!

julianrkung (changes requested):
Some small changes! You are honestly the best I am about to cry ;-;

src/screens/Home/Home.tsx
jenmhoang: We might need to access currentSite from the Home page in order to display the correct summarized numbers for the selected site. Alternatively, I can grab sites from Home and pass it to SiteMenu to maintain consistency

src/screens/Home/components/HomeMenuItem.tsx
jenmhoang: also agree that `amount` doesn't seem to fit, but not sure what else would be better. thoughts on `alerts` or `itemAlerts`? also not in love with those ideas either though

src/screens/Home/Home.tsx
jenmhoang: :'''''''''''' ahhhhhhhhhh

src/screens/Home/components/SiteMenu.tsx
jenmhoang: it looks like MaterialUI's `MenuItem` component uses a deprecated React component -- they're apparently planning on resolving this issue in version 5 of MaterialUI (which is in alpha), and this warning apparently only shows up in StrictMode development (won't show an error in production build). I think it's fine for now, but we can try to check back on this later if needed (and potentially just replace the `MenuItem` MaterialUI component with our own styled component)

src/screens/Home/components/HomeMenuItem.tsx
julianrkung: I can make a note of it and keep thinking, but this shouldn't block. the PR from merging. I approved, thank you :)

julianrkung (approved):
Thank u thank u thank u

src/screens/Home/Home.tsx
kyleqhua: ^^^ I think its fine that its pulled here as we will use it later.

kyleqhua (commented):
All the components were so well done omg. No significant changes/issues I saw. I think when we integrate functionality with this, it shouldn't be that big of a deal. Just adding a couple lines to the Home.tsx

kyleqhua (approved):

PR #27, opened 2020-12-28
Moved API calls to meter readings and payments to backend

julianrkung:
**Note**: File is blocked by https://github.com/calblueprint/meepanyar/pull/26 . Do not merge until this is unblocked.

**Changelist**:
- Moved API calls for customers / meter readings / payments to backend (used to be in frontend). Associated PR for backend: https://github.com/calblueprint/meepanyar-node/pull/3 
- Added a patch and associated README that should be run after new `airtable-schema-generator` runs that changes `customerIds` to `customers` on the site schema
- `refreshSiteData` no longer takes in parameters because it never used the parameters in the first place.


src/lib/airtable/patches/README.md
```diff
@@ -0,0 +1,10 @@
+### Overview
```
jenmhoang, line 1: this is so neat you are the best

jenmhoang: Should we also add preservations of non-generated functions in request.js that would be overwritten upon re-generation?

src/lib/airtable/schema.js
```diff
@@ -29,7 +29,7 @@ export const Columns = {
 	"Sites": {
 		name: {name:`Name`, type:`text`},
 		userIds: {name:`Users`, type:`foreignKey-many`},
-		customerIds: {name:`Customers`, type:`foreignKey-many`},
+		customers: {name:`Customers`, type:`custom-object`}, // Modified via patch.
 		financialSummarieIds: {name:`Financial Summaries`, type:`foreignKey-many`},
 		tariffPlanIds: {name:`Tariff Plans`, type:`foreignKey-many`},
 		field6: {name:`Field 6`, type:`text`},
```
jenmhoang, line 35: looks like an unnecessary extra field was added into the Airtable and included in the generation

jenmhoang (approved):
your genius is unmatched

src/lib/airtable/patches/README.md
julianrkung: Good idea. I'll do that in another PR that'll also remove the extra airtable field you included in the other comment

src/lib/airtable/schema.js
julianrkung: Good catch. I'll remove it in another PR that'll also add a patch for the non-generated functions

PR #28, opened 2020-12-30
Cleaned up env file and service worker uses environment file for network endpoint

julianrkung:
Service worker used to catch failed network requests to a hardcoded URL endpoint. This PR changes that so that the service worker catches network requests to a URL endpoint specified by environment variables.

The transfer of environment variables was necessary because environment variables don't seem to be available to service workers in build environments. To remedy this, we manually transfer the environment variables of interest by writing to a JS file called `sw-env.js` during the `npm build` process.

**Note**: It's too out-of-the-way to properly test this in this PR (we require being able to send POST requests and making sure they're sent when back online). This PR doesn't test for that functionality. Instead, we will test that when PRs related to POST requests start landing (some time early next week).

### Testing PR
- run `npm run build` in the root directory
- Check that the resulting `build` file produced properly transfers the variables in the `.env` file into `build/sw-env.js`.
- Check that service workers still work when running `npm run build`
  - After building, run `serve -s build` and log in
  - Check for errors with service worker in devtools console
  - Turn off your network connection via devtools, cancel the `serve -s build` process
  - Go to `localhost:5000` and check that you still arrive a the home page

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,4 @@
+AIRTABLE_EMAIL= # Needed when using the schema generator
+AIRTABLE_PASSWORD= # Needed when using the  schema generator
+REACT_APP_AIRTABLE_ENDPOINT_URL= # Point towards airlock server
 REACT_APP_AIRTABLE_BASE_ID=
-REACT_APP_AIRTABLE_API_KEY=
-# REACT_APP_AIRTABLE_ENDPOINT_URL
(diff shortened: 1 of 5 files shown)


kyleqhua (approved):
Followed instructions as written. Works as expected with the exception that only http://localhost:5000 is able to load. Any routes in the browser will cause app to error. Went over with @julianrkung

PR #29, opened 2021-01-04
Regenerate Airtable schema and updated patch README

julianrkung:
**Changelist**:
- Redid patch system documentation / framework so that it produces less conflicts & better clarity.
- Regenerated schema to remove excess fields
- Added patch to add all the non-generated functions and fix spelling errors
- URL fields no longer hardcoded and now use the environment variable `REACT_APP_AIRTABLE_ENDPOINT_URL`
- Removed fields `createManyPayments` and `createManyMeterReadingsandInvoices` because they won't be used and not worth the confusion for onboarding devs

**Testing**
- Try branching off of main, regenerating the schema, and applying the patches in this PR as instructed and see if you run into any conflicts.
- Maybe also see how sensitive applying patches is when fields are added. I added a field in "Customer Updates" and it didn't cause a patch conflict, but its worth trying again to see if there's a workflow within reason that might break it.


src/lib/airtable/patches/README.md
```diff
@@ -1,10 +1,21 @@
 ### Overview
 This folder contains patches that should be run after regenerating the airtable schema. This is needed to maintain consistency between the frontend and backend.
 
-Add additional patches here via `git format-patch ... -o <path/to/dir>`.
-For more information, you can view the resource [here](https://devconnected.com/how-to-create-and-apply-git-patch-files/)
+### Overview
+This folder contains patches that should be run after regenerating the airtable schema. This is needed to maintain consistency between the frontend and backend.
```
wangannie: I'm guessing this was duplicated by accident?

src/lib/airtable/patches/README.md
```diff
@@ -1,10 +1,21 @@
...
+
+## Add Additional Patches
```
wangannie, line 10: UX nit: maybe swap this with **Directions to apply** since I'm assuming that part is most relevant?

src/lib/airtable/patches/patch-01.patch
```diff
@@ -0,0 +1,266 @@
+diff --git a/src/lib/airtable/request.js b/src/lib/airtable/request.js
```
wangannie, line 1: I don't know much about patches overall so feel free to ignore this but would it be helpful to rename this file to something more descriptive, esp if there could potentially be multiple patches for different purposes?

wangannie (approved):
Tested with the steps you described in a new branch off main following the README and I didn't run into any issues! Updating naming + cleanup looks good too. I just left a few small notes on formatting/patch naming.

Re: adding a field, I tested that as well on a few different tables and field types, and applying the patch wasn't an issue as long as I ran the schema generator each time. If I add a field without updating schema, then applying the patch was rejected (which I'm guessing is expected behavior).

src/lib/airtable/patches/README.md
julianrkung: Thank you yes mb

src/lib/airtable/patches/patch-01.patch
julianrkung: This naming convention was chosen so that patches are run in the correct order (in the case of dependent patches). Hopefully this shouldn't be too much of an issue since there shouldn't be many patches here. Maybe it'd be good to document exactly what each patch does somewhere in the doc. I'll leave that as a future todo, thanks!

PR #30, opened 2021-01-20
Changed route names to be kebab-case and work with offline

julianrkung:
Changed route names to be more CRUD-like (kebab-case and ending verbs).

Route have also been changed to not depend on `rid`. This is because users who are created offline do not receive an `rid` until they return online and are created on the Airtable. Therefore, we can't create routes based on `rid` because not all customers have it. Instead, because we store customer information anyways for offline use, we have a generic `/customers/customer` route that we can pass customer information to populate.

### To Test
- Ensure that all changed routes are still navigable (functionality preserved)
- Check convention

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -37,17 +37,17 @@ function App(isSignedIn: AppProps) {
       <AuthenticatedRoute path="/home" component={Home} />
 
       <AuthenticatedRoute path="/customers" component={CustomerMain} exact />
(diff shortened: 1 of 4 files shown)


wangannie (approved):
I wasn't too familiar with what functionality was already there (adding meter readings isn't working when I test this?), but other existing functionality seems normal + conventions look good!

julianrkung:
Yeah, functionality hasn't been implemented yet for actually adding meter readings.  Future sprint! Thanks for the review!

PR #34, opened 2021-02-17
Add Customer FAB

wangannie:

# What's new in this PR

Added a floating action button to the Customer screen to add a new customer.

## How to review

- Make sure it works as expected and aligns with Figma

### Online sources

- [FAB API docs from material-ui](https://material-ui.com/api/fab/)
- [FAB demo](https://material-ui.com/components/floating-action-button/#animation) (from which I pretty much copy-pasted the styling).

### Related PRs

n/a

## Next steps

### Screenshots

image

CC: @julianrkung

### DIFF
--- a/src/screens/Customers/CustomerMain.tsx
+++ b/src/screens/Customers/CustomerMain.tsx
@@ -1,13 +1,15 @@
+import { createStyles, Fab, FormControl, FormHelperText, MenuItem, Select, Theme, withStyles } from '@material-ui/core';
+import AddIcon from '@material-ui/icons/Add';
 import React, { useEffect, useState } from 'react';
-import { CustomerRecord } from '../../lib/airtable/interface';
 import { Link, RouteComponentProps } from 'react-router-dom';
-import { store } from '../../lib/redux/store';
-import { createStyles, FormControl, FormHelperText, MenuItem, Select, Theme, ThemeProvider, withStyles } from '@material-ui/core';
-import UserSearchBar from '../../components/UserSearchBar';
-import CustomerCard from '../../components/CustomerCard';
-import TrieTree from '../../lib/utils/TrieTree';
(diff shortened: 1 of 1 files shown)


src/screens/Customers/CustomerMain.tsx
```diff
@@ -100,6 +108,9 @@ function CustomerMain(props: RouteComponentProps & UserProps) {
           </FormControl>
         </div>
       </div>
+      <Fab href="/customers/create" color='primary' aria-label='add customer' className={classes.fab} size='medium'>
+        <AddIcon fontSize="large"/>
+      </Fab>
```
julianrkung: We've been using `Link` components. Unless there's a good reason to switch off, let's keep this the standard. You can read about the difference [here](https://stackoverflow.com/questions/43087007/react-link-vs-a-tag-and-arrow-function). There are some examples of it being used throughout the codebase (like in this file).

A practical reason we're doing this is because non-Links cause refreshes (from the link) which can interrupt in-progress http requests (which we have to account for due to offline func since http requests can end up firing at any point the user is navigating the app)

julianrkung (commented):
Looks pretty much good, just small changes!

Also: The chevrons seem to be above the FAB rather than below it, and so there's some unexpected interaction when scrolling (see screenshot)

image

Also the button turns blue when clicked which we probably want to avoid (It looks awkward, idk what it should look like when held down though)

image

src/screens/Customers/CustomerMain.tsx
wangannie: Got it. This also somehow fixed the issue where it was turning blue, so now it has the proper animations on click! 

Also good catch on the chevron poking over the button! I fixed it by moving the FAB below the customer cards.

julianrkung (approved):
Wonderful. This makes me happy. Thank you.

PR #35, opened 2021-02-23
Customer Overview Sorting and Filtering

micahtyong:

# What's new in this PR

Closes issue described [here](https://www.notion.so/Customer-Overview-sort-feature-f7567cc56efe4785a6dc470fb9d63282). In summary, we add support for sorting and filtering results in the customer overview page. 

## How to review

1. [Setup repository locally](https://github.com/calblueprint/meepanyar/blob/main/README.md). 
2. Run `npm start`. Navigate to Customer Tab (2nd to the left). 
3. Select your sort by and filter by attributes. 

### Online sources
- [Multiselect from MaterialUI](https://material-ui.com/components/selects/#multiple-select)

### Screenshots
Users can now navigate to the customer overview screen and see their results sorted alphabetically and grouped by active status by default. In the following example, results are grouped by meter status.  
Screen Shot 2021-02-23 at 5 16 39 PM

Users can specify how to sort and filter the customer results by using this multi-select dropdown. 
Screen Shot 2021-02-23 at 11 48 03 AM

CC: @julianrkung


src/screens/Customers/CustomerMain.tsx
```diff
@@ -37,25 +37,72 @@ interface UserProps {
...
-  }, []);
+enum SortBy {
+  name = 'Name (A - Z)' as any, 
+  meter = 'Meter Number' as any
+}
+
+enum FilterBy {
+  paymentStatus = 'Payment Status' as any, 
+  meterStatus = 'Meter Status' as any,
+  activeStatus = 'Active Status' as any,
```
julianrkung: enum property naming convention is usually all capital with underscores, so let's stick to that!

`NAME`, `PAYMENT_STATUS`, etc

(as a side note I do think that all the `any` definitions are weird, but I can't figure out a good way to circumvent it since I'm not that familiar with Typescript, so I'm willing to just be okay with it for now)

src/screens/Customers/CustomerMain.tsx
```diff
@@ -37,25 +37,72 @@ interface UserProps {
...
+    allCustomers.sort(sortByFunction());
+    allCustomers.filter(filterByFunction);
     setFilteredCustomers(allCustomers);
     setFullCustomers(allCustomers);
     const customersTrie: TrieTree<CustomerRecord> = new TrieTree('name');
     customersTrie.addAll(allCustomers);
     setAllCustomersTrie(customersTrie);
+  }
 
+  const sortByFunction = (): (a: CustomerRecord, b: CustomerRecord) => number => {
```
julianrkung: Can you comment this function / rename it so that it's clear that this function chooses a sorting function to sort the users on rather (and isn't a sorting function itself)? The name makes it sound like the function we use to sort the data which can be a bit misleading (since this is what the `filterByFunction` does)

julianrkung: Also an extra set of parentheses might make this type definition more readable to typescript novices like me. Took me a while to finally interpret it as 

`(): ((a: CustomerRecord, b: CustomerRecord) => number) => {...}`

I also would like to mention that this is a really clean way to implement, good work!

src/screens/Customers/CustomerMain.tsx
```diff
@@ -83,23 +130,43 @@ function CustomerMain(props: RouteComponentProps & UserProps) {
     setFilteredCustomers(customers);
   }
 
+  const handleMenuSelect = (e: any) => {
+    const keys: string[] = e.target.value; 
+    const key: string = keys[keys.length - 1];
```
julianrkung: I think this naming convention might be a bit confusing for people reading the code after the fact. Maybe we could be more specific with these variable names?

like `selectedOptions` and `newlySelectedOption` or something along those lines? Otherwise I think it becomes unclear why the last key is taken from the list (it also might be good to comment why the last key is chosen for those who look at this code later and wonder)

src/screens/Customers/CustomerMain.tsx
```diff
@@ -83,23 +130,43 @@ function CustomerMain(props: RouteComponentProps & UserProps) {
     setFilteredCustomers(customers);
   }
 
+  const handleMenuSelect = (e: any) => {
```
julianrkung: Kinda bad on my part because i let the `handleSearchChange` function merge with an `any` parameter type, but MaterialUI types these the following way (i took this typing from one of their sample code), so let's try to follow suit?

image

src/screens/Customers/CustomerMain.tsx
```diff
@@ -83,23 +130,43 @@ function CustomerMain(props: RouteComponentProps & UserProps) {
...
+    if (sortByKeys.includes(key)) {
+      setSortBy(SortBy[key as keyof typeof SortBy]);
+      return; 
+    }
+    
+    const filterByKeys: string[] = Object.keys(FilterBy); 
+    if (filterByKeys.includes(key)) {
+      setFilterBy(FilterBy[key as keyof typeof FilterBy]);
+      return; 
+    }
```
julianrkung: Really clean! I like it :)

src/screens/Customers/CustomerMain.tsx
```diff
@@ -37,25 +37,72 @@ interface UserProps {
...
+      case (SortBy.name): {
+        return (a: CustomerRecord, b: CustomerRecord) => { return a.name.localeCompare(b.name) }
+      }
+      default: {
+        return (a: CustomerRecord, b: CustomerRecord) => { return 0 }
+      }
+    }
+  }
+
+  const filterByFunction = (e: CustomerRecord): boolean => {
```
julianrkung: NIT: change to `customer` instead of `e` for variable name

julianrkung (commented):
Really clean stuff, code was easy to read :). Most comments are about naming convention.

I also noticed that the frontend is a bit off from the Figma? I think it's missing the delineation after choosing a filter (the records get split into Paid / Unpaid, Meter / No Meter, and Active / Inactive when a filter is chosen). Can we get that in for this sprint? Screenshot below for that.

image

src/screens/Customers/CustomerMain.tsx
micahtyong: Yeah, it's the only way I was able to reverse map values to key without having to iterate through the enum. Gonna keep my eye out on that one.

src/screens/Customers/CustomerMain.tsx
micahtyong: Gotcha, I'll write a comment for this.

src/screens/Customers/CustomerMain.tsx
```diff
@@ -78,39 +88,53 @@ function CustomerMain(props: RouteComponentProps & UserProps) {
...
     switch (filterBy) {
-      case (FilterBy.meterStatus): {
-        return e.hasmeter
+      case (FilterBy.METER_STATUS): {
+        return customer.hasmeter
       }
-      case (FilterBy.paymentStatus): {
-        return e.payments.length > 0
+      case (FilterBy.PAYMENT_STATUS): {
+        return customer.payments.length > 0
```
julianrkung: I should clarify with Ashley exactly how this should be sorted (whether it's if people have paid this period or if it's related to outstanding balance).

Can we change this so that it filters customers that have a zero outstanding balance & people with a non-zero outstanding balance? The payments array is related to all the payments the customer has made in their lifetime as a customer.

julianrkung (commented):
1 small change and 1 search frontend thing. Other than that it looks really good, we can merge after the 2 things are addressed, good stuff!

Search bug:
Looks like when I try to search it starts at a weird point in the bar. Was that caused by this PR? If not, we can address it in another PR. Screenshot below so you see what i'm talking about.

image

src/screens/Customers/CustomerMain.tsx
micahtyong: Gotcha! After meeting w Ashley, your intuition is right that it's regarding outstanding balance. My next commit will include this

src/screens/Customers/CustomerMain.tsx
```diff
@@ -159,12 +163,13 @@ function CustomerMain(props: RouteComponentProps & UserProps) {
...
-    }
-    const customers = allCustomersTrie.get(searchVal);
-    sortAndFilterCustomers(customers);
+    setSearchValue(searchVal); 
+    // if (searchVal === '') {
+    //   sortAndFilterCustomers(fullCustomers); 
+    //   return;
+    // }
+    // const customers = allCustomersTrie.get(searchVal);
+    // sortAndFilterCustomers(customers);
```
julianrkung: delete commented code

julianrkung (approved):
LGTM, go ahead and merge after deleting the commented code.

Really good work! I learned a lot of typescript while reviewing this haha

PR #36, opened 2021-02-24
Kyle/unpaid

kyleqhua:

# What's new in this PR
Add unpaid report calculations on the home screen and imported my old calculations over. Will edit more later

- calculates unpaid reports: finds the amount of reports where billed > paid
- calculates to charge: finds the amount of customers that still need to be charged
- calculates outstanding payments: finds the amount of customers that still owes money.

home.tsx
momentUtil.js

Uses Tiffany's fin sum prs.

### Screenshots
Screen Shot 2021-03-03 at 4 17 36 PM
Screen Shot 2021-03-03 at 4 18 16 PM

CC: @julianrkung


src/screens/Home/Home.tsx
```diff
@@ -37,6 +38,70 @@ interface HomeProps {
...
 function Home(props: HomeProps) {
   const { classes, lastUpdated, isOnline, currentSite } = props;
+  useEffect(() => {
+    proccessData();
+  }, []);
+
+  const [outStandingPayments, setOutstandingPayments] = useState<number>();
+  const [toCharge, setToCharges] = useState<number>();
+  const [customerDetails, setCustomerDetails] = useState<number>();
+  const [unpaidReports, setUnpaidReports] = useState<number>();
```
julianrkung: I don't think this has to be a stateful component. We can calculate the data on render and show it without using state

src/screens/Home/Home.tsx
```diff
@@ -37,6 +38,70 @@ interface HomeProps {
...
+    setToCharges(toCharge);
+    setCustomerDetails(toCharge + outstanding);
+  }
+
+  //true has outstadining false has paid
+  const checkPayment = (customer: CustomerRecord) => {
+    if (parseInt(customer.outstandingBalance) > 0) {
+      return true;
+    }
+    return false;
```
julianrkung: reduce to 1 line. `return parseInt(customer.outstandingBalance) > 0`

src/screens/Home/Home.tsx
```diff
@@ -37,6 +38,70 @@ interface HomeProps {
...
+  }
+
+  const calculateUnpaidReports = (ss: SiteRecord) => {
```
julianrkung: What does ss stand for? Can we just name it `site`?

src/screens/Home/Home.tsx
```diff
@@ -37,6 +38,70 @@ interface HomeProps {
...
+  const [unpaidReports, setUnpaidReports] = useState<number>();
+
+
+
+  const proccessData = () => {
+    calculateCustomerData(currentSite);
+    calculateUnpaidReports(currentSite)
+  }
+
+  const calculateCustomerData = (ss: SiteRecord) => {
```
julianrkung: What does ss stand for? Can we just name it `site`?

src/screens/Home/Home.tsx
```diff
@@ -37,6 +38,70 @@ interface HomeProps {
...
+    // customers to charge:  haven't charged yet/ no meterreading
+    // outstanding payments: done the meter reading / charged, haven't paid yet
+    let toCharge = 0;
+    let outstanding = 0;
+    //get day
+    const today = new Date();
+    const currMonth = today.getUTCMonth();
```
julianrkung: We should use the moment library for date handling. It'd be good to write whatever util functions you need in `momentUtils` to add a layer of abstraction. See Tiffany's PR https://github.com/calblueprint/meepanyar/pull/21 for a good example

src/screens/Home/Home.tsx
```diff
@@ -37,6 +38,70 @@ interface HomeProps {
...
+    let customers = ss.customers;
+    for (let i = 0; i < customers.length; i++) {
+      let curr = customers[i]
+      //depends if the meter readings list is sorted with earliest => latest
+      let latestMeterReading = curr.meterReadings[curr.meterReadings.length - 1];
+      if (latestMeterReading != null) {
+        const dateTime = Date.parse(latestMeterReading.date);
```
julianrkung: I don't think it's sorted in redux. We shouldn't impose that invariant arbitrarily since it just gets pulled and stored directly from Airtable.

src/screens/Home/Home.tsx
```diff
@@ -37,6 +38,70 @@ interface HomeProps {
...
+  const proccessData = () => {
+    calculateCustomerData(currentSite);
+    calculateUnpaidReports(currentSite)
+  }
+
+  const calculateCustomerData = (ss: SiteRecord) => {
+    // customers to charge:  haven't charged yet/ no meterreading
+    // outstanding payments: done the meter reading / charged, haven't paid yet
+    let toCharge = 0;
+    let outstanding = 0;
```
julianrkung: `outstandingPayments` is a better variable name. Outstanding is weird because it's an actual word with a main definition that isn't relevant to what it's doing

src/screens/Home/Home.tsx
```diff
@@ -37,6 +38,70 @@ interface HomeProps {
...
+    //get day
+    const today = new Date();
+    const currMonth = today.getUTCMonth();
+    let customers = ss.customers;
+    for (let i = 0; i < customers.length; i++) {
+      let curr = customers[i]
```
julianrkung: Make variable `const` if it's not changing. Also choose a more descriptive variable name. `currCustomer` or something like that to show what the object is

src/screens/Home/Home.tsx
```diff
@@ -37,6 +38,70 @@ interface HomeProps {
...
+      //depends if the meter readings list is sorted with earliest => latest
+      let latestMeterReading = curr.meterReadings[curr.meterReadings.length - 1];
+      if (latestMeterReading != null) {
+        const dateTime = Date.parse(latestMeterReading.date);
+        let readingDate = new Date(dateTime);
+        if (readingDate.getUTCMonth() < currMonth) {
```
julianrkung: Write a util function in the moment utils file so we have a standardized way of date comparison. The function could pass in the string dates we get from Airtable and do proper comparisons

src/screens/Home/Home.tsx
```diff
@@ -37,6 +38,70 @@ interface HomeProps {
 
 function Home(props: HomeProps) {
   const { classes, lastUpdated, isOnline, currentSite } = props;
+  useEffect(() => {
+    proccessData();
+  }, []);
+
+  const [outStandingPayments, setOutstandingPayments] = useState<number>();
+  const [toCharge, setToCharges] = useState<number>();
```
julianrkung: `numOutstandingPayments` and `numCustomersToCharge` aren't too long and make this a lot easier to read imo

src/screens/Home/Home.tsx
```diff
@@ -54,15 +119,15 @@ function Home(props: HomeProps) {
       <Link to={'/customers'}>
         <HomeMenuItem
           label="Customer Alerts"
-          amount={1}
+          amount={customerDetails ? customerDetails : 0}
           sublabels={[
             { amount: 1, label: 'Customers to Charge' },
             { amount: 2, label: 'Outstanding Payments' },
```
julianrkung: Change the amount in these sublabels to show the correct numbers from your calculations

julianrkung (commented):
Left some comments!

src/screens/Home/Home.tsx
```diff
@@ -38,6 +39,56 @@ interface HomeProps {
 function Home(props: HomeProps) {
   const { classes, lastUpdated, isOnline, currentSite } = props;
 
+  const calculateCustomerData = () => {
+    // customers to charge:  haven't charged yet/ no meterreading
+    // outstanding payments: done the meter reading / charged, haven't paid yet
+    let toCharge = 0;
+    let outstandingPayments = 0;
+    let customers = currentSite.customers;
```
julianrkung: sorry I think my feedback before wasn't the best on this. I think the clearest variable name is `numCustomersToCharge` and `numOutstandingPayments`.

`customers` should also be a `const` since it doesn't change.

src/screens/Home/Home.tsx
```diff
@@ -38,6 +39,56 @@ interface HomeProps {
...
+          outstandingPayments += 1;
+        }
+      } else {
+        toCharge += 1;
+      }
+    }
+    console.log('payments')
+    console.log(outstandingPayments);
+    console.log('need to charge')
+    console.log(toCharge);
```
julianrkung: remove logs

src/screens/Home/Home.tsx
```diff
@@ -38,6 +39,56 @@ interface HomeProps {
...
+    let total = outstandingPayments + toCharge;
+    return { 'outstandingPayments': outstandingPayments, 'toCharge': toCharge, 'totalAmount': total }
```
julianrkung: We should just do the addition of `totalAmount` when passing the `amount` prop to the component. `amount={customerData.oustandingPayments + customerData.toCharge}` is clearer because a `totalAmount` variable has no context/history in the app / screen other than what this function arbitrarily assigns to it which makes it unintuitive for a developer reading the code.

src/screens/Home/Home.tsx
```diff
@@ -38,6 +39,56 @@ interface HomeProps {
...
+    let unpaid = 0;
+    let summaries = currentSite.financialSummaries;
+    for (let i = 0; i < summaries.length; i++) {
+      let singleSummary = summaries[i];
+      if (singleSummary.totalCustomersBilled > singleSummary.totalCustomersPaid) {
+        unpaid += 1;
+      }
+    }
+    console.log('unpaid')
+    console.log(unpaid)
```
julianrkung: remove logs

src/screens/Home/Home.tsx
```diff
@@ -38,6 +39,56 @@ interface HomeProps {
...
+  }
+
+  //true has outstadining false has paid
+  const checkPayment = (customer: CustomerRecord) => {
+    return parseInt(customer.outstandingBalance) > 0;
+  }
+
+  const calculateUnpaidReports = () => {
+    let unpaid = 0;
+    let summaries = currentSite.financialSummaries;
```
julianrkung: const

src/screens/Home/Home.tsx
```diff
@@ -38,6 +39,56 @@ interface HomeProps {
...
+    return { 'outstandingPayments': outstandingPayments, 'toCharge': toCharge, 'totalAmount': total }
+  }
+
+  //true has outstadining false has paid
+  const checkPayment = (customer: CustomerRecord) => {
+    return parseInt(customer.outstandingBalance) > 0;
+  }
+
+  const calculateUnpaidReports = () => {
+    let unpaid = 0;
```
julianrkung: `numUnpaidReports`

src/screens/Home/Home.tsx
```diff
@@ -38,6 +39,56 @@ interface HomeProps {
...
+    // outstanding payments: done the meter reading / charged, haven't paid yet
+    let toCharge = 0;
+    let outstandingPayments = 0;
+    let customers = currentSite.customers;
+    for (let i = 0; i < customers.length; i++) {
+      const currCustomer = customers[i]
+      //depends if the meter readings list is sorted with earliest => latest
+      let latestMeterReading = currCustomer.meterReadings[currCustomer.meterReadings.length - 1];
+      if (latestMeterReading != null) {
+        let lessMonth = lessThanCurrentMonth(latestMeterReading.date)
```
julianrkung: const.

Also we should look for a better name here. `isReadingFromCurrentPeriod` or something. The `if` statement below should ideally make intuitive sense as a standalone statement for someone reading it.

src/lib/moment/momentUtils.ts
```diff
@@ -5,3 +5,14 @@ import moment from 'moment';
 export const formatUTCDateStringToLocal = (dateStringUTC: string): string => {
   return moment(dateStringUTC).local().format('YYYY-MM-DD HH:mm:ss');
 };
+
+//takes in a string of a date time YYYY-MM-DD format. Returns true if its before the current month.
+export const lessThanCurrentMonth = (dateString: string): boolean => {
+  let lastDate = moment(dateString);
+  let lastMonth = lastDate.get('month');
+  let currentMonth = moment().get('month');
```
julianrkung: const

src/lib/moment/momentUtils.ts
```diff
@@ -5,3 +5,14 @@ import moment from 'moment';
...
+  if (lastMonth < currentMonth) {
+    return true;
+  }
+  return false;
```
julianrkung: you can just do `return lastMonth < currentMonth` instead of an if else that returns true and false

src/screens/Home/Home.tsx
```diff
@@ -38,6 +39,50 @@ interface HomeProps {
...
+  const calculateCustomerData = () => {
+    // customers to charge:  haven't charged yet/ no meterreading
+    // outstanding payments: done the meter reading / charged, haven't paid yet
+    let toCharge = 0;
+    let outstandingPayments = 0;
+    let customers = currentSite.customers;
+    for (let i = 0; i < customers.length; i++) {
+      const currCustomer = customers[i]
+      //depends if the meter readings list is sorted with earliest => latest
+      let latestMeterReading = currCustomer.meterReadings[currCustomer.meterReadings.length - 1];
```
julianrkung: Don't think we can depend on the meterReadings list being sorted from earliest to latest (that's not an invariant established yet in our model, but we can). We either need to enforce that "sorted" invariant from the backend or go through all the meter readings and make sure we grab the latest one here. 

Also `const` for `latestMeterReading` since it doesn't change.

src/screens/Home/Home.tsx
```diff
@@ -38,6 +39,50 @@ interface HomeProps {
...
+      if (latestMeterReading != null) {
+        let lessMonth = lessThanCurrentMonth(latestMeterReading.date)
+        if (lessMonth) {
+          toCharge += 1;
+        }
+        if (checkPayment(currCustomer)) {
```
julianrkung: This 1-liner I think makes the code harder to read. I don't know what "checking" a payment means from this context (and I have to look into the function to figure it out). We should either improve the function name so that it's clear what this `if` statement actually checks, or, since the function is a 1-liner, is simple, and only used here, we could remove the function and do the logic inside of the if statement (separating it into 2 lines for clarity would help).

src/screens/Home/Home.tsx
```diff
@@ -38,6 +39,50 @@ interface HomeProps {
...
+          outstandingPayments += 1;
+        }
+      } else {
+        toCharge += 1;
+      }
+    }
+    let total = outstandingPayments + toCharge;
```
julianrkung: const

src/screens/Home/Home.tsx
```diff
@@ -38,6 +39,50 @@ interface HomeProps {
...
+    return { 'outstandingPayments': outstandingPayments, 'toCharge': toCharge, 'totalAmount': total }
+  }
+
+  //true has outstadining false has paid
+  const checkPayment = (customer: CustomerRecord) => {
+    return parseInt(customer.outstandingBalance) > 0;
+  }
+
+  const calculateUnpaidReports = () => {
```
julianrkung: We should ask Ashely, but I'm pretty sure unpaid reports in this context are reports that the user hasn't paid off, not that there are customers left to collect from.

The user needs to have gone through the below payment screens to have their report considered "paid" i think. 

We could possibly add a `Paid` field to see if they've paid or we could check for a copy of the bankslip. I don't know what to do for part-payments, though. I'll ask the NPO tomorrow if that's a concern.

image

src/lib/moment/momentUtils.ts
```diff
@@ -5,3 +5,14 @@ import moment from 'moment';
 export const formatUTCDateStringToLocal = (dateStringUTC: string): string => {
   return moment(dateStringUTC).local().format('YYYY-MM-DD HH:mm:ss');
 };
+
+//takes in a string of a date time YYYY-MM-DD format. Returns true if its before the current month.
+export const lessThanCurrentMonth = (dateString: string): boolean => {
+  let lastDate = moment(dateString);
+  let lastMonth = lastDate.get('month');
```
julianrkung: `lastMonth` and `lastDate` aren't good variable names here since they hint that it's from a month ago

src/lib/moment/momentUtils.ts
```diff
@@ -5,3 +5,14 @@ import moment from 'moment';
 export const formatUTCDateStringToLocal = (dateStringUTC: string): string => {
   return moment(dateStringUTC).local().format('YYYY-MM-DD HH:mm:ss');
 };
+
+//takes in a string of a date time YYYY-MM-DD format. Returns true if its before the current month.
+export const lessThanCurrentMonth = (dateString: string): boolean => {
```
julianrkung: Function doesn't consider january --> december. We need to consider periods rather than months (months aren't enough info). Tiffany merged some helpful moment functions if you want to pull (you might not even need to create a function if you're able to do the logic properly with Tiffany's existing util functions).

kyleqhua: oh I didnt thake this into consideration ty

src/lib/airtable/schema.js
```diff
@@ -1,6 +1,6 @@
 /*
-    THIS IS A GENERATED FILE
-    Changes might be overwritten in the future, edit with caution!
+	THIS IS A GENERATED FILE
```
julianrkung: Can you revert the formatting changes made here? We don't want to change the generated file apart from apply patches, or else we risk patch conflicts (because they depend on diffs).

src/lib/moment/momentUtils.ts
```diff
@@ -6,6 +6,18 @@ export const formatUTCDateStringToLocal = (dateStringUTC: string): string => {
   return moment(dateStringUTC).local().format('YYYY-MM-DD HH:mm:ss');
 };
 
+
+export const lessThanCurrentPeriod = (dateString: string): boolean => {
+  const currentDate = moment(dateString);
+  const currentMonth = moment().startOf('month');
+  const currentDatePeriod = currentDate.valueOf();
+  const currentPeriod = currentMonth.valueOf();
```
julianrkung: Naming is confusing here. It makes it especially hard to follow because `currentMonth` refers to the actual current month, but `currentDate` refers to the date given after parsing the date string (which could be past dates). 

We need a better name than `currentDate` and etc so that the logic is easy to read and understand

src/lib/moment/momentUtils.ts
```diff
@@ -6,6 +6,18 @@ export const formatUTCDateStringToLocal = (dateStringUTC: string): string => {
...
+  if (currentDatePeriod < currentPeriod) {
+    return true;
+  }
+  return false;
```
julianrkung: instead of `if (...) return true else false` just return the evaluation of the if statement

src/screens/Home/Home.tsx
```diff
@@ -1,11 +1,12 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
```
julianrkung: Remove unused imports

src/screens/Home/Home.tsx
```diff
@@ -38,6 +39,45 @@ interface HomeProps {
...
+    let numCustomersToCharge = 0;
+    let numOutstandingPayments = 0;
+    const customers = currentSite.customers;
+    for (let i = 0; i < customers.length; i++) {
+      const currCustomer = customers[i]
+      //depends if the meter readings list is sorted with earliest => latest
+      let latestMeterReading = currCustomer.meterReadings[currCustomer.meterReadings.length - 1];
+      if (latestMeterReading != null) {
+        let lessMonth = lessThanCurrentPeriod(latestMeterReading.date);
+        if (lessMonth) {
```
julianrkung: https://github.com/calblueprint/meepanyar/pull/36#discussion_r586927473

You can put the function call inside the `if` statement since the variable is only used once if thinking of a good name is difficult here

julianrkung (commented):
More comments! Good improvement :) .

julianrkung (commented):
I think we can merge after the unresolved comments are taken care of

src/lib/airtable/airtable.js
```diff
@@ -177,7 +177,7 @@ function getRecordsByAttribute(
       view: VIEW,
       filterByFormula: filterByFormula
         ? `AND(${filterByFormula}, {${fieldType}}=${fieldValue})`
-        : `{${fieldType}}=${fieldValue}`,
+        : `({${fieldType}}="${fieldValue}")`,
```
julianrkung: Can this be deleted? Was this a mistake

julianrkung (approved):
Looks good! :) We can merge and work on revamping the home page now

PR #38, opened 2021-03-03
Tiffany/payment payslip

cnnmon:

# What's new in this PR

- Front end of the financial summary/payslip screen
- Takes in financial record as a prop & takes/displays amount billed
- Button to go to camera screen

## How to review

http://localhost:3000/confirm-payment
Doesn't take in the right props currently as the screen leading into it (?) isn't built out yet.

## Next steps

- Camera functionality
- Passing down the right props
- Sending data to Airtable

### Screenshots

image

CC: @julianrkung


src/screens/FinancialSummary/ConfirmPayment.tsx
```diff
@@ -0,0 +1,78 @@
...
+      height: '41px',
+      width: '70%',
+      marginTop: 15,
+    },
+    cameraButton: {
+      backgroundColor: '#F7F9FC',
+      width: '95%',
+      height: 130,
+      margin: 10,
+      border: '3.5px dashed ' + theme.palette.divider,
```
julianrkung: NIT
```
border: `3.5px dashed ${theme.palette.divider}`,
```
template literal here is cleaner to read

src/screens/FinancialSummary/components/FinancialInfo.tsx
```diff
@@ -0,0 +1,90 @@
...
+      fontWeight: 700,
+      color: theme.palette.primary.main,
+    },
+    fieldContainer: {
+      marginTop: '10px',
+    },
+  }),
+);
+
+interface CardProps {
```
julianrkung: `FinancialInfoProps`

src/screens/FinancialSummary/components/PaymentCard.tsx
```diff
@@ -0,0 +1,77 @@
...
+);
+
+interface CardProps {
+  payment: number;
+  paid: number;
+}
+
+export default function PaymentCard(props: CardProps) {
+  const classes = useStyles();
+  const getPaymentButtons = () => {
```
julianrkung: singular since it's 1 button

src/screens/FinancialSummary/components/PaymentCard.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
```
julianrkung: Is this component used anywhere? I can't match this with anything on the [current Figma ](https://www.figma.com/file/MzpMtoYyVYFn19DpAXHEci/Mee-Panyar?node-id=2161%3A2).

If it's not used anywhere, we might wanna not have this in the PR. A good rule of thumb is that we only PR things that are being used and are testable by the reviewer

src/screens/FinancialSummary/components/FinancialInfo.tsx
```diff
@@ -0,0 +1,90 @@
+import React from 'react';
+import { Typography, Button, Divider } from '@material-ui/core';
```
julianrkung: Remove unused import

src/screens/FinancialSummary/ConfirmPayment.tsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import { Link, RouteComponentProps } from 'react-router-dom';
```
julianrkung: Remove unused import

src/App.tsx
```diff
@@ -44,6 +45,7 @@ function App(isSignedIn: AppProps) {
       <AuthenticatedRoute path="/customers/customer/meter-readings/create" component={AddMeterReading} exact />
       <AuthenticatedRoute path={'/customers/customer/records'} component={CustomerRecords} exact />
       <AuthenticatedRoute path="/financial-summaries" component={FinancialSummariesMain} />
+      <AuthenticatedRoute path="/confirm-payment" component={ConfirmPayment} />
```
julianrkung: We'll want to change this route once we've built out the remainder of the Financial Summary screens. It'll eventually look something like `/financial-summaries/financial-summary/payment/create` imo. Can we add a TODO above this to change the route name once the other screens are built out?

src/screens/FinancialSummary/ConfirmPayment.tsx
```diff
@@ -0,0 +1,78 @@
...
+    }
+});
+
+interface ConfirmPaymentProps extends RouteComponentProps {
+  classes: { content: string; confirmButton: string; cameraButton: string; cameraText: string; };
+  financialSummary: FinancialSummaryRecord;
+  location: any;
+}
+
+function ConfirmPayment(props: ConfirmPaymentProps) {
```
julianrkung: Name isn't descriptive enough. It doesn't tell enough about what this component is (ideally, i'd be able to match a specific component like this one to the screen it's on just by looking at it's name). For components especially, longer names are better (unless it's egregiously long). I wouldn't mind something along the lines of `FinancialSummaryPayment` (which isn't really perfect either)

julianrkung (commented):
@cnnmon To make it easier for reviewers (so it doesn't error), could we for now just use the first Financial Summary when we go to the route (also add a TODO to remove this once the screen that leads into this is built out)? That's my bad since for some reason I thought Maria had built out the component that links into this one the semester before.

I added the following to `ConfirmPayment` (and imported the functions at the top) as a quick hack to get it to run without error FYI @wangannie when you review.

```
  financialSummary: state.siteData.currentSite.financialSummaries[0] || null,
});

export default connect(mapStateToProps)(withStyles(styles)(ConfirmPayment));

```

src/screens/FinancialSummary/components/PaymentCard.tsx
```diff
@@ -0,0 +1,77 @@
...
+      color: 'white',
+      backgroundColor: theme.palette.primary.main,
+    },
+    unbolded: {
+      fontWeight: 500,
+    },
+  }),
+);
+
+interface CardProps {
```
wangannie: Rename to be more specific

src/screens/FinancialSummary/ConfirmPayment.tsx
```diff
@@ -0,0 +1,78 @@
...
+      letterSpacing: '0.14em',
+      height: '41px',
+      width: '70%',
+      marginTop: 15,
+    },
+    cameraButton: {
+      backgroundColor: '#F7F9FC',
+      width: '95%',
+      height: 130,
+      margin: 10,
```
wangannie: Nit to match Figma: this is supposed to fit flush with the rest of the text labels, so you don't need this margin here. Just for best practice, I'd also cut the `width` here and instead add this to the `content` container:
```
display: 'flex',
flexDirection: 'column',
```

wangannie: I see there's also a 10 margin in `FinancialInfo` -- I'd also take that out to make it flush.

src/screens/FinancialSummary/ConfirmPayment.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <BaseScrollView>
+        <div className={classes.content}>
+          <FinancialInfo bankName={"Aurora"} accountNumber={203} accountName={"Mee"} balance={financialSummary.totalAmountBilled} />
+          <Typography variant="h2" color="textPrimary">
+            Add Photo of Payslip
+          </Typography>
+          <Button className={classes.cameraButton} variant="contained" color="primary" disableElevation={true}>
+            <div className={classes.cameraText}>
+              <PhotoLibraryIcon /><br />
+              Upload Photo
```
wangannie: For the sake of consistency, you should probably wrap this in a Typography tag. Then you don't have to use the `<br />` and you can use flexbox to put these in a centered column layout!

src/screens/FinancialSummary/components/FinancialInfo.tsx
```diff
@@ -0,0 +1,90 @@
...
+    right: {
+      textAlign: 'right',
+    },
+    bolded: {
+      fontWeight: 700,
+    },
+    divider: {
+      margin: '10px 0px',
+    },
+    balanceText: {
```
wangannie, line 30: Looks like this should be red according to Figma

src/screens/FinancialSummary/components/FinancialInfo.tsx
```diff
@@ -0,0 +1,90 @@
...
+            <Typography className={classes.bolded} variant="body1">
+              Account Number
+            </Typography>
+            <Typography className={classes.bolded} variant="body1">
+              Account Name
+            </Typography>
+          </div>
+          <div className={classes.right}>
+            <Typography variant="body1">{props.bankName}</Typography>
+            <Typography variant="body1">**** **** {props.accountNumber}</Typography>
```
wangannie: Is this ****** temporary? I'm curious how sensitive this number is/how we should handle this.

wangannie (commented):
I didn't dive too deep into the CSS but I left some higher-level styling comments. @julianrkung mentioned styling isn't the highest priority right now, so defer to him on if you should address/ignore my comments. Otherwise, this looks great! A+

src/screens/FinancialSummary/components/FinancialInfo.tsx
julianrkung: True, I think on the Adalo app it showed the entire accounting number (so it might be easier to just do that). I don't think it's a big deal either way, though.

src/screens/FinancialSummary/FinancialSummaryPayment.tsx
```diff
@@ -0,0 +1,90 @@
...
+    },
+    header: {
+      marginTop: 5,
+      marginBottom: 10,
+    },
+});
+
+interface FinancialSummaryPaymentProps extends RouteComponentProps {
+  classes: { content: string; confirmButton: string; cameraButton: string; header: string; };
+  financialSummaries: FinancialSummaryRecord[]; //TODO: change later to take in one financial summary
```
julianrkung: Is there a reason this TODO can't be done right now? Looks like we can just pass the `EMPTY_SUMMARY` if we don't have a summary in redux.

src/lib/redux/siteDataSlice.ts
```diff
@@ -25,6 +25,22 @@ export const EMPTY_SITE: SiteRecord = {
   tariffPlans: [],
 };
 
+export const EMPTY_FINANCIAL_SUMMARY: FinancialSummaryRecord = {
+  name: '',
```
julianrkung, line 30: add `id: '',` to top of this since `FinancialSummaryRecord` now has ids!

image

julianrkung (approved):
LGTM after you add the `id` field to `EMPTY_FINANCIAL_SUMMARY`.

Good work :) high quality PR!

PR #6, opened 2020-10-21
Navigation

jenmhoang:
original PR by @emmastephan (moved branches)
> Added in bottom tab navigation with icons that change color when clicked on, and a header with a hamburger icon that opens a menu that slides out from the side.

PR #39, opened 2021-03-03
Add tariff info to customer profile

wangannie:

# What's new in this PR

- Added the tariff info section to the top of the customer profile
  - I just adjusted the `OutlinedCardList` to have a `columns` prop (not sure if this is a very intuitive name since technically the whole cardlist becomes a row...which is the opposite..anyways I'm open to suggestions) which basically switches it to lay the stats side by side.
  - If, for any reason, a tariff value is blank/invalid, it should show `-`
- Minor styling updates to bring it closer to the designs -- final front-end changes are on hold pending finalized designs.

## How to review

Run the `annie/tariff-plans` in the node app ([Node PR #6](https://github.com/calblueprint/meepanyar-node/pull/6)) to get the updated tariff info.

### Online sources

https://material-ui.com/styles/basics/#adapting-based-on-props

### Related PRs

https://github.com/calblueprint/meepanyar-node/pull/6

## Next steps
- Update/finalize styling!

### Screenshots

**Normal behavior**
image

**Edge case: missing/invalid values**
image

CC: @julianrkung


src/components/OutlinedCardList.tsx
```diff
@@ -1,6 +1,19 @@
...
+  label: string;
+  unit: string;
+}
+
+interface CardProps {
+  info: CardPropsInfo[];
+  primary: boolean;
+  rightIcon?: JSX.Element;
+  columns?: boolean;
+}
```
julianrkung, line 16: Is there a reason this was moved to be above the styles? If not can we move it back to preserve blame history?

src/lib/airtable/interface.ts
```diff
@@ -46,8 +47,8 @@ export interface TariffPlanRecord {
 export interface CustomerRecord {
   name: string;
   meterNumber: number;
-  tariffPlansId: string[];
-  tariffPlans: TariffPlanRecord[];
+  tariffPlansId: string;
+  tariffPlans: TariffPlanRecord;
```
julianrkung, line 51: Yeah this naming is annoying and confusing I'll change it during worksession tonight, sorry!

julianrkung (approved):
LGTM! Thanks for getting this done, good work :)

I wasn't too stringent on the styling because designs aren't finalized, but if this design is what we end up with, we should standardize the tariff portion (IMO it's offsetting that sometimes the `MMK` units are on a 3rd line and sometimes on the 2nd line)

src/components/OutlinedCardList.tsx
wangannie: Hm I thought I had to move this because I use the props in `makeStyles`

julianrkung: It still compiles and runs if i move it back (I think interfaces might be hoisted?). 

Actually, it's probably better practice to just put all types at the top (more important than preserving blame in this case). I'm fine with this diff, sorry!

wangannie:
Good catch! Updated to move unit to a new line when it's in this column layout.
image

PR #43, opened 2021-03-04
Add customer functionality

wangannie:

# What's new in this PR

- The Add Customer page is fully functional -- it should add a customer to Redux even when offline and then add to Airtable.
- Updated the `TextField` and `Checkbox` base components to include `onChange` so we can start making them functional everywhere else!
- Bugfix: switched navigating to customer profiles from `CustomerMain` to use `/customers/customer` instead of names

## How to review

- When offline, add a customer and make sure it shows up in both the Customers list frontend and in Redux
- When online, add a customer and make sure it does all of the above + shows up in Airtable!

### Online sources

https://material-ui.com/components/selects/
https://stackoverflow.com/questions/122102/what-is-the-most-efficient-way-to-deep-clone-an-object-in-javascript

### Related PRs

n/a

## Next steps
- See TODO comments!
- Handle edge cases and add input validation/error states in the form (pending finalized designs)
- Update the frontend match the designs
- Un-nest/refactor customer data (see [comment](https://github.com/calblueprint/meepanyar/pull/43#discussion_r588080971))

### Screenshots

CC: @julianrkung


src/components/Button.tsx
```diff
@@ -15,15 +15,17 @@ const styles = () =>
     },
   });
 
+  // TODO: @wangannie figure out how onclick would work in relation to linking to avoid confusion
 interface MainButtonProps {
   classes: { button: string };
   label: string;
+  onClick?: (event: React.MouseEvent) => void;
```
wangannie, line 22: I need this `onClick` for the `handleSubmit` on the add customer form, but I know that in most other places, these buttons are used with `<Link>` tags instead, so I worry about this being confusing. Maybe there's a way to consolidate to `onClick` and wrap the Button in `Link` tags here, then just pass in the destination through `onClick`? Or I'm overthinking it and this is fine :~)

src/components/Checkbox.tsx
```diff
@@ -29,10 +29,13 @@ const styles = (theme: Theme) =>
     },
   });
 
+  // TODO: @wangannie look into renaming the onChange props
```
wangannie: Since we need 2 onChange functions this is how I ended up naming them, but I'm open to suggestions!

src/components/TextField.tsx
```diff
@@ -14,20 +14,27 @@ const styles = (theme: Theme) =>
     },
   });
 
+  // TODO: @wangannie: make onChange not optional
```
wangannie, line 17: I  _think_ eventually I need to make `onChange` not optional, but I'd have to go through all the other existing `TextFields` and add `onChange` functions that do nothing for now.

src/lib/redux/siteData.ts
```diff
@@ -25,4 +25,8 @@ const setCurrentSite = (newSite: any): void => {
   store.dispatch(setCurrSite(newSite));
 };
 
-export { refreshSiteData, setCurrentSite };
+const addCustomer = (customer: any): void => {
```
wangannie: Is this the right place to have this? It doesn't feel super related to site (I guess it's just adding a customer to the current site?).

src/lib/redux/siteData.ts
```diff
@@ -25,4 +25,8 @@ const setCurrentSite = (newSite: any): void => {
...
+  store.dispatch(addNewCustomer(customer));
```
wangannie: I'm a little iffy on my naming -- I was looking at the example of `setCurrSite` and `setCurrentSite` and assumed it'd be good to use separate names, so I have `addCustomer` and `addNewCustomer`. But we also do `createCustomer`, so I'm not sure if this naming is more confusing / if I should be more explicit that it's to redux??

src/lib/redux/siteDataSlice.ts
```diff
@@ -41,8 +41,19 @@ const siteDataSlice = createSlice({
     setCurrSite(state, action) {
       state.currentSite = action.payload;
     },
+    addNewCustomer(state, action) {
```
wangannie: This nesting was based on the [article you linked](https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns#updating-nested-objects) -- it's definitely  not the cleanest way, but I decided not to dig into `createReducer` last night so if you think we should still switch to that and/or un-nest, lmk.

src/screens/Customers/AddCustomer.tsx
```diff
@@ -1,38 +1,116 @@
...
+  const handleSubmit = (event: React.MouseEvent) => {
+    // Prevent page refresh on submit
+    event.preventDefault();
+
+    const customer = EMPTY_CUSTOMER;
+    customer.name = customerName;
+    customer.meterNumber = parseInt(meter);
+    customer.isactive = !customerInactive;
+    customer.tariffPlansId = selectedTariffPlan;
+    addCustomer(customer);
```
wangannie: More on my iffy naming [in this comment](https://github.com/calblueprint/meepanyar/pull/43/files#r587639748), but here it looked especially sketchy :~) especially with `customer` and `cust`...

wangannie: Also, existing customer records in redux usually do have a `sitesId` field, but you mentioned adding the basic `CustomerRecord` version into Redux and then tacking on other relevant info to send to Airtable. That means the redux version would still be missing a few fields -- is that okay? Would it cause problems later on bc not all customer records have the same set of attributes?

src/screens/Customers/EditCustomer.tsx
```diff
@@ -28,11 +28,11 @@ function EditCustomer(props: EditCustomerProps) {
   return (
     <BaseScreen title="Edit Customer" leftIcon="backNav">
       <form noValidate className={classes.content}>
-        <TextField label={'Name'} />
+        <TextField label={'Name'} id={'name'}/>
```
wangannie, line 31: FYI I didn't work on this screen -- I just had to add `id`s when I merged in `main` since I made it a required prop. If you were wondering, we need `id` for accessibility!

src/screens/Customers/AddCustomer.tsx
```diff
@@ -1,38 +1,116 @@
...
+
 function AddCustomer(props: AddCustomerProps) {
-  const { classes } = props;
+  const { classes, currentSite } = props;
+
+  const [selectedTariffPlan, setSelectedTariffPlan] = useState("");
+  const [customerName, setCustomerName] = useState("");
+  const [meter, setMeter] = useState("");
+  const [meterChecked, setMeterChecked] = useState(false);
+  const [customerInactive, setCustomerInactive] = useState(false);
```
wangannie: Not sure how much this really matters, but do you think I should rename these to match the interface/schema? Or does that not matter?

src/components/TextField.tsx
julianrkung: I agree. Don't do it in this PR. Save for a dedicated PR

src/lib/redux/siteDataSlice.ts
julianrkung: IMO it'd be cleanest to unnest. That way we can have a `customerDataSlice` where it'd make sense to put `EMPTY_CUSTOMER` (might be better in utils but `EMPTY_SITE` is in `siteDataSlice`) and `addCustomer` makes more sense in that file. 

This might be a lot of refactor work, so maybe I could do it in a subsequent PR.

**NOTE**: We'd have to be careful, though when changing sites. We need to make sure that if a User is in charge of 2 sites, `Site A` and `Site B`, that the scenario where:
1.  I create a customer to `Site A`, 
2. Switch to `Site B` 
3. Switch back to `Site A`
that the customer I created in `Site A` is still visible (we can do this by persisting changes to `customers` into `currentSite` whenever a site change is made)

src/screens/Customers/AddCustomer.tsx
```diff
@@ -1,38 +1,116 @@
...
-import { RouteComponentProps } from 'react-router-dom';
+import Checkbox from '../../components/Checkbox';
+import TextField from '../../components/TextField';
+import { CustomerRecord, SiteRecord } from '../../lib/airtable/interface';
+import { createCustomer } from '../../lib/airtable/request';
+import { addCustomer } from '../../lib/redux/siteData';
+import { RootState } from '../../lib/redux/store';
+
+// TODO: Move to utils file?
+const EMPTY_CUSTOMER : CustomerRecord= {
```
julianrkung: Move to utils file would be good. This is potentially useful for tests once they're set up in the future

src/screens/Customers/AddCustomer.tsx
julianrkung: Might be very slightly better but as long as the code is readable (which in this case it is) it doesn't matter much. I wouldn't change it.

In fact, changing `customerInactive` to `IsActive` would probably make the code harder to read because then the checkbox would need to be negated, so I like how it is now.

src/screens/Customers/AddCustomer.tsx
julianrkung: Yeah, we should standardize. It'd be better to just cut out any unnecessary fields before sending the data to the frontend. Ex: we won't need `sitesId` insides our customers object ever, so there's no reason to keep it in the customer objects we send and store in redux

src/screens/Customers/AddCustomer.tsx
```diff
@@ -1,38 +1,116 @@
...
+    const customer = EMPTY_CUSTOMER;
+    customer.name = customerName;
+    customer.meterNumber = parseInt(meter);
+    customer.isactive = !customerInactive;
+    customer.tariffPlansId = selectedTariffPlan;
+    addCustomer(customer);
+    
+    const cust = {
+      ...customer,
+      sitesId: currentSite.rid
```
julianrkung: NIT: comment an explanation on why this is needed. It might also be cleaner to just not store it as a variable (since then you'd have to think of a name) and instead just call the `createCustomer` function and do the object instantiation on the same line.

```
createCustomer({
...customer,
sitesId: currentSite.rid
});
```

src/screens/Customers/EditCustomer.tsx
julianrkung: Yeah i think there was some merge weirdness with one of Tiffany's PRs a while back, thanks for putting it back!

src/components/Checkbox.tsx
julianrkung: Naming is fine here, I like it!

src/components/Checkbox.tsx
```diff
@@ -1,8 +1,8 @@
-import React from 'react';
-import { withStyles, createStyles, Theme } from '@material-ui/core/styles';
-import { TextField, Checkbox } from '@material-ui/core';
+import { Checkbox, TextField } from '@material-ui/core';
```
julianrkung, line 1: Can you also add a `TODO` to rename this component? I was confused why this checkbox had a text field also. I don't know a good name off the top of my head though. Will think.

src/components/Button.tsx
julianrkung: I think this is fine but i'll take another look in the morning (brain is kinda fried)

julianrkung (commented):
Addressed your comments, I'll look at the actual content of the PR tomorrow. I think this is very close to be able to merge, so might just have you finish it up and I can do the redux un-nesting in another PR.

Could you update the `next steps` part of the description so that it's up to date & accurate? I'm going to use it to keep track of what needs to happen after this PR.

src/components/Button.tsx
julianrkung: update: should be fine. if it becomes a problem we can revisit

src/lib/redux/siteData.ts
julianrkung: I think `addCustomerToRedux` might be good since `addCustomer` might be confused with adding to Airtable. I should've named `setCurrSite` to `setCurrentSite` in hindsight. Don't know why I abbreviated

src/lib/redux/siteData.ts
julianrkung: can you add a TODO for me to move this into a `customerDataSlice`? I agree, doesn't look like it belongs here

src/screens/Customers/AddCustomer.tsx
```diff
@@ -1,38 +1,116 @@
...
   location: any;
+  currentSite: SiteRecord;
 }
 
+
 function AddCustomer(props: AddCustomerProps) {
-  const { classes } = props;
+  const { classes, currentSite } = props;
+
+  const [selectedTariffPlan, setSelectedTariffPlan] = useState("");
```
julianrkung: name should mention that it's the `tariffPlanId` and not the tariffPlan.

src/screens/Customers/AddCustomer.tsx
julianrkung: Is there an associated backend change? `currentSite` doesn't have an `rid` field right now on my frontend / backend. If there is an associated backend change, would you mind linking to it?

image

src/screens/Customers/AddCustomer.tsx
```diff
@@ -1,38 +1,116 @@
...
+
+  const handleSelectTariffPlan = (event: React.ChangeEvent<{ value: unknown }>) => {
+    setSelectedTariffPlan(event.target.value as string);
+  }
+
+  const handleSubmit = (event: React.MouseEvent) => {
+    // Prevent page refresh on submit
+    event.preventDefault();
+
+    const customer = EMPTY_CUSTOMER;
```
julianrkung: We want to make a copy of `EMPTY_CUSTOMER` and mutate that. Otherwise, subsequent `EMPTY_CUSTOMER`s will have some fields already filled out if we submit a 2nd customer

src/screens/Customers/AddCustomer.tsx
```diff
@@ -1,38 +1,116 @@
...
 }
 
+
 function AddCustomer(props: AddCustomerProps) {
-  const { classes } = props;
+  const { classes, currentSite } = props;
+
+  const [selectedTariffPlan, setSelectedTariffPlan] = useState("");
+  const [customerName, setCustomerName] = useState("");
+  const [meter, setMeter] = useState("");
```
julianrkung: TODO: I should check what constraints to put on meters (numeric, alphanumeric, etc...). I don't see a meter number even being set in the current Adalo app, so I should look into it.

src/screens/Customers/AddCustomer.tsx
```diff
@@ -1,38 +1,116 @@
...
+          <InputLabel id="select-tariff-plan-label">Select Tariff Plan</InputLabel>
+          <Select label={"Select Tariff Plan"} id={'select-tariff-plan'} labelId = "select-tariff-plan-label" onChange={handleSelectTariffPlan}>
+            {tariffPlans.map((plan, index) =>
+                <MenuItem key={index} value={plan.id}>{plan.name}</MenuItem>
+            )}
+          </Select>
+        </FormControl>
+        <Checkbox label={'Meter:'} textField={meterChecked ? 'meter': null} checkboxOnChange={() => setMeterChecked(!meterChecked)}textFieldOnChange={handleMeterInput}/>
+        <Checkbox label={'Customer is inactive'} checkboxOnChange = {() => setCustomerInactive(!customerInactive)}/>
+        <Button label={'Add'} onClick={(event) => handleSubmit(event)} />
```
julianrkung: just `onClick={handleSubmit}` works, no?

src/screens/Customers/AddCustomer.tsx
```diff
@@ -1,38 +1,116 @@
...
-        <Checkbox label={'Meter:'} textField={'meter'} />
-        <TextField label={'Tariff Plan'} />
-        <Button label={'ADD'} />
+      <form noValidate className={classes.content} onSubmit={() => false}>
+        <TextField label={'Name'} id={'name'} primary={true} onChange={handleNameInput} />
+        <FormControl variant="outlined" className={classes.formControl}>
+          <InputLabel id="select-tariff-plan-label">Select Tariff Plan</InputLabel>
+          <Select label={"Select Tariff Plan"} id={'select-tariff-plan'} labelId = "select-tariff-plan-label" onChange={handleSelectTariffPlan}>
+            {tariffPlans.map((plan, index) =>
+                <MenuItem key={index} value={plan.id}>{plan.name}</MenuItem>
```
julianrkung: For best practice, index is a bad key. Something unique for each item that's being mapped is best. We can use `plan.id` as key i think.

Source: [here](https://medium.com/wesionary-team/using-index-as-a-key-is-an-anti-pattern-in-react-8e5db3aea3a9)

src/screens/Customers/AddCustomer.tsx
```diff
@@ -1,38 +1,116 @@
...
+        <Checkbox label={'Customer is inactive'} checkboxOnChange = {() => setCustomerInactive(!customerInactive)}/>
+        <Button label={'Add'} onClick={(event) => handleSubmit(event)} />
       </form>
     </BaseScreen>
   );
 }
 
-export default withStyles(styles)(AddCustomer);
+const mapStateToProps = (state: RootState) => ({
+  currentSite: state.siteData.currentSite
```
julianrkung: Right now, `currentSite` is defaulted to none (I'm going to change this in a PR to just default to EMPTY_SITE). There shouldn't be a case where it ends up null here, but better to write defensive code for now in case that changes. Can you do `... || EMPTY_SITE`?

src/screens/Customers/AddCustomer.tsx
```diff
@@ -1,38 +1,116 @@
...
 
+
 function AddCustomer(props: AddCustomerProps) {
-  const { classes } = props;
+  const { classes, currentSite } = props;
+
+  const [selectedTariffPlan, setSelectedTariffPlan] = useState("");
+  const [customerName, setCustomerName] = useState("");
+  const [meter, setMeter] = useState("");
+  const [meterChecked, setMeterChecked] = useState(false);
```
julianrkung: Not entirely happy with the `meterChecked` name since it might imply that the customer's meter has been checked. Maybe `hasMeter`?

julianrkung (commented):
Looks really good! 1 quick iteration should be enough to finish. Picked a bunch of small naming convention things.

Bigger: Is there an associated backend PR? If so, would you mind adding that as a PR on the node repo and linking it in the PR's description?

src/lib/redux/siteData.ts
wangannie: Renamed this one but leaving the other one as `addCustomer` just because it will yell at me if I try 'Import declaration conflicts with local declaration'

src/screens/Customers/AddCustomer.tsx
wangannie: Oh beautiful

src/screens/Customers/CustomerMain.tsx
```diff
@@ -221,7 +221,7 @@ function CustomerMain(props: RouteComponentProps & UserProps) {
         }
         <FormHelperText>{filterLabels[1]}</FormHelperText>
         {filteredCustomersAlt.map((customer, index) => (
-          <Link key={index} to={{ pathname: `${props.match.url}/${customer.name}`, state: { customer: customer } }}>
+          <Link key={index} to={{ pathname: `${props.match.url}/customer`, state: { customer: customer } }}>
```
julianrkung, line 224: For future devs reading: this change was a bug missed in a previous PR that's addressed here. The rationale for a `/customer` endpoint is listed in the description of this commit https://github.com/calblueprint/meepanyar/commit/daa912a21b950e315227756fa661da2030832d40

src/screens/Customers/AddCustomer.tsx
```diff
@@ -49,42 +33,47 @@ interface AddCustomerProps extends RouteComponentProps {
...
+    setSelectedTariffPlanId(event.target.value as string);
   }
 
+  // TODO: Add form input validation and error messaging
   const handleSubmit = (event: React.MouseEvent) => {
     // Prevent page refresh on submit
     event.preventDefault();
 
-    const customer = EMPTY_CUSTOMER;
+    const customer = JSON.parse(JSON.stringify(EMPTY_CUSTOMER));
```
julianrkung, line 55: NIT: Add a comment explaining that this is to copy the empty customer object, otherwise line looks kinda weird

src/screens/Customers/AddCustomer.tsx
```diff
@@ -49,42 +33,47 @@ interface AddCustomerProps extends RouteComponentProps {
...
-    const cust = {
+    // Add other info necessary to create the Airtable record
+    createCustomer({
       ...customer,
-      sitesId: currentSite.rid
-    }
-    
-    createCustomer(cust);
+      siteId: currentSite.id
+    });
```
julianrkung: Can you put a TODO for error handling? We're fine with letting network connection errors occur (since app is supposed to work offline), but we should catch other errors when creating customers and adding to redux

src/lib/utils/customerUtils.ts
```diff
@@ -57,3 +57,21 @@ export const getTariffPlan = (customer: CustomerRecord, currentSite: SiteRecord)
...
 }
+
+export const EMPTY_CUSTOMER : CustomerRecord = {
+  id: '',
+  name: '',
+  meterNumber: 0,
+  tariffPlanId: '',
+  isactive: false,
+  hasmeter: false,
+  outstandingBalance: '',
```
julianrkung: should be 0 instead of empty string. Otherwise the Total Owed doesn't show 0 in the customer main like it does with other customers. 

image

julianrkung (commented):
1 quick fix and a couple TODO / NITs and then will be good to merge after

julianrkung (approved):
LGTM! Thanks, it's nice we're getting functionality in :)

PR #52, opened 2021-03-14
InventoryData setup + Add Inventory Functionality

wangannie:

# What's new in this PR
- Added setup for the `inventoryData` slice.
- Basic inventory setup -- the main list of inventory works with cards that link to (mostly empty) inventory profiles
- Working functionality to add an inventory item (aka add a product to a site). The list of options only shows the products that aren't already in the site

## How to review
To test the add functionality, run `annie/inventory-items` on the backend ([companion PR](https://github.com/calblueprint/meepanyar-node/pull/16))
- Navigate to the inventory tab (shopping cart)
- Click the **+** button
- Try adding inventory and see if it:
  - limits selection to products that the site doesn't already have
  - adds the new inventory record to redux
  - shows up in Airtable correctly
  - navigates to the new profile page corresponding to the correct inventory item + qty
  
Note: if there aren't any choices in the dropdown, it may be because the site already carries every product, so you'll have to delete them from Airtable. I think we need to eventually address this edge case through the designs.
  
**Switching between sites**
- In `testUser`'s account, select the East Asian Library site and add an inventory item
- Switch to Kyle's dancing places and check that the inventory list has changed
- Add an inventory item to Kyle's dancing places and check that it shows up properly.

### Online sources

- https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype
- https://material-ui.com/components/autocomplete/

### Related PRs
Corresponding backend PR: https://github.com/calblueprint/meepanyar-node/pull/16

## Next steps
- Automatically trigger an inventory update when creating a new inventory item
- Add input validation and error messaging
- Add the ability to add new products
- Add relevant details to inventory profile + inventory card to match designs
- Clean up/remove data we don't use (e.g. `inventoryIds` in `currentSite` and products) 

### Screenshots

image
image

CC: @julianrkung


src/lib/airtable/request.js
```diff
@@ -12,18 +12,21 @@
...
   createRecords,
-  updateRecord,
-  updateRecords,
-  getAllRecords,
-  getRecordsByAttribute,
-  getRecordById,
-  deleteRecord,
+
+
+
```
julianrkung: Some weird formatting here. I'd also prefer to not mess with the order of imports in the generated files. I think this is done automatically via your linter, but you can do `CMD+K S` to save without formatting (I usually do `CMD+SHIFT+P` and just search for the "save without formatting" command but i just noticed there's a shortcut for it)

src/lib/redux/inventoryData.ts
```diff
@@ -0,0 +1,39 @@
...
+import { getAllProducts, getInventorysByIds, getInventoryUpdatesByIds, getPurchaseRequestsByIds } from '../airtable/request';
+import { saveInventoryData, setLoadingForInventoryData } from './inventoryDataSlice';
+import { store } from './store';
+
+const refreshInventoryData = async (currentSite: SiteRecord): Promise<void> => {
+  store.dispatch(setLoadingForInventoryData());
+  if (currentSite) { // TODO @wangannie what to do if site null
+    const products = await getAllProducts();
+
+    const currentSiteInventory = await getInventorysByIds(currentSite.inventoryIds);
```
julianrkung: Is there a reason you're making the extra request on the frontend instead of gathering the inventory data on the backend when pulling sites (aka pulling all the inventory data inside of the siteData access resolver)?

If there's good grounds for it, that's fine, but if not I'd prefer it done in the `Sites` access resolver on the backend because

1. It's faster (1 API call instead of 2)
2. I'm a little weary about trusting the internet connection for Users being flaky. I'd rather do the majority of API requests on the backend as a safety measure since the backend is under a more reliable environment.

If you have reasons for the frontend, that's fine just wanted to let you know the justification for the current layout

src/lib/redux/inventoryData.ts
```diff
@@ -0,0 +1,39 @@
...
+    
+    // TODO @wangannie: look into avoiding .flat, simplifying/eliminating .filter
+    const inventoryWithPurchaseRequests = currentSiteInventory.filter((inventory: InventoryRecord) => inventory.purchaseRequestIds);
+    let purchaseRequests: PurchaseRequestRecord[] = [];
+    await Promise.all(inventoryWithPurchaseRequests.map(async (inventory: InventoryRecord) => 
+      getPurchaseRequestsByIds(inventory.purchaseRequestIds)
```
julianrkung: We generally wanna minimize API requests whenever possible. Wouldn't it be faster to gather all the `purchaseRequestIds` for all inventory, make 1 call to `getPurchaseRequestsByIds` and then do matching of `purchaseRequests` to their inventory afterwards?

src/lib/redux/inventoryData.ts
```diff
@@ -0,0 +1,39 @@
...
+    )).then(data => purchaseRequests = data as PurchaseRequestRecord[]);
+    purchaseRequests = purchaseRequests.flat(1);
+
+      
+    const inventoryWithUpdates = currentSiteInventory.filter((inventory: InventoryRecord) => inventory.inventoryUpdateIds);
+    let inventoryUpdates: InventoryUpdateRecord[] = [];
+    await Promise.all(inventoryWithUpdates.map(async (inventory: InventoryRecord) => 
+      getInventoryUpdatesByIds(inventory.inventoryUpdateIds)
```
julianrkung: Same comment as ^ with the minimize API requests whenever possible and/or moving them to the backend

src/lib/redux/inventoryData.ts
```diff
@@ -0,0 +1,39 @@
...
+    )).then(data => inventoryUpdates = data as InventoryUpdateRecord[]);
+    inventoryUpdates = inventoryUpdates.flat(1);
```
julianrkung: I think `flat` defaults to 1?

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable no-unused-vars */
+/* eslint-disable no-param-reassign */
+import { createSlice } from '@reduxjs/toolkit';
+import { InventoryRecord, InventoryUpdateRecord, PurchaseRequestRecord } from '../airtable/interface';
+
+interface inventoryDataSliceState {
+  isLoading: boolean;
+  products: any[];
```
julianrkung: Is this not supposed to be a `ProductRecord[]`?

src/screens/Inventory/AddInventory.tsx
```diff
@@ -0,0 +1,93 @@
...
+    formControl: {
+      width: '100%',
+    },
+  });
+
+interface AddInventoryProps extends RouteComponentProps {
+  classes: { content: string, formControl: string};
+  location: any;
+  currentSite: SiteRecord;
+  products: any[];
```
julianrkung: Is this not supposed to be a ProductRecord[]?

src/App.tsx
```diff
@@ -52,6 +52,7 @@ function App(isSignedIn: AppProps) {
       <AuthenticatedRoute path="/confirm-payment" component={FinancialSummaryPayment} />
 
       <AuthenticatedRoute path="/inventory" component={Inventory} />
+      <AuthenticatedRoute path="/inventory/create" component={AddInventory} />
```
julianrkung: I think the error you were talking about when navigating to `/inventory/create` is caused by the lines here. The path `/inventory/create` triggers both of these I think since `/inventory/create` is a suffix of `/inventory`. To fix, we can put `exact` in both routes so `/inventory` won't trigger on `/inventory/create`

src/screens/Inventory/AddInventory.tsx
```diff
@@ -0,0 +1,93 @@
...
+    event.preventDefault();
+
+    // Make a deep copy of an empty inventory record
+    const inventory = JSON.parse(JSON.stringify(EMPTY_INVENTORY));
+    inventory.productId = selectedProductId;
+    inventory.siteId = currentSite.id;
+    inventory.currentQuantity = startingAmount;
+    inventory.periodStartQuantity = startingAmount;
+
+    createInventory(inventory);
```
julianrkung: Like the AddCustomer functionality, we also want to save separately to redux in case of offline. This can be another PR if you want, but it'd be good to put a TODO in that case

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -0,0 +1,60 @@
...
+  siteInventory: InventoryRecord[];
```
julianrkung: We might want to do some restructuring later on how the data is stored. I think we want to easily be able to access the data for each site, so storing within a map based on `siteId` might be a good call. I think this is generally a problem with our current `siteData` structure also, which I need to fix.

The inefficiency I see with this is when we modify the redux store, we'd have to do a deep copy of all inventory (products, purchase requests, etc...) across all the sites in the redux store when we should only really need to make a deep copy of the inventory in the site where we made the inventory modification.

Again, not a huge deal right now and we might be able to do a data restructure some time (maybe spring break) if you wanna put it off until then. Sorry for putting off the `customerData` refactor  so late so you didn't have a reference

julianrkung (commented):
Some feedback (and the reason for the bug you put in the description)! Sorry for the late review

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -0,0 +1,68 @@
...
+  updatedQuantity: 0,
+  inventoryId: '',
+  period: 0,
+};
+
+const inventoryDataSlice = createSlice({
+  name: 'inventoryData',
+  initialState,
+  reducers: {
+    setLoadingForInventoryData(state) {
```
julianrkung: I don't think this is necessary anymore right?

src/lib/redux/siteData.ts
```diff
@@ -17,7 +18,14 @@ const refreshSiteData = async (): Promise<void> => {
...
   };
-
+  // Extract to inventoryData
+  refreshInventoryData(currentSite);
+
+  // Clear inventory fields from currentSite
+  delete currentSite.products;
+  delete currentSite.inventory;
+  delete currentSite.inventoryUpdates;
+  delete currentSite.purchaseRequests;
```
julianrkung: Shouldn't this happen for all sites?

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -0,0 +1,68 @@
...
+interface inventoryDataSliceState {
+  isLoading: boolean;
+  products: ProductRecord[];
+  siteInventory: InventoryRecord[];
+  purchaseRequests: PurchaseRequestRecord[];
+  inventoryUpdates: InventoryUpdateRecord[];
+}
+
+const initialState: inventoryDataSliceState = {
+  isLoading: false,
```
julianrkung: `isLoading` was initially necessary because we sent API request and had to wait for it to be fulfilled. I think this isn't necessary anymore because we no longer send API request (we just use `siteData`'s)

src/lib/redux/siteData.ts
julianrkung: When you `setCurrentSite` doesn't your `inventorySlice` need to change also?

julianrkung (commented):
Quick PR comments. Talk offline!

src/lib/redux/inventoryDataSlice.ts
julianrkung: { SiteIdA : {
purchaseRequests,
...
},
SiteIdB: {
purchaseRequests,
...
}

Products: []
}

In siteDataSlice: {

currentSiteId: SiteIdB
}

src/lib/airtable/schema.js
```diff
@@ -45,6 +45,10 @@ export const Columns = {
 		tariffPlanIds: {name:`Tariff Plans`, type:`foreignKey-many`},
 		id: {name:`ID`, type:`formula`},
 		inventoryIds: {name:`Inventory`, type:`foreignKey-many`},
+		products: { name: `Products`, type: `custom-object` },
+		inventory: { name: `SiteInventory`, type: `custom-object` },
```
julianrkung: Ping me before we merge and we can update the patch for `schema.js` so these stay on regeneration

src/lib/redux/inventoryData.ts
```diff
@@ -0,0 +1,28 @@
+import { SiteRecord } from '../airtable/interface';
+import { addInventory, saveInventoryData } from './inventoryDataSlice';
+import { store } from './store';
+
+const refreshInventoryData = async (currentSite: SiteRecord): Promise<void> => {
```
julianrkung: Is there a reason this is async?

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const EMPTY_SITE_INVENTORY_DATA : SiteInventoryData = {
+  siteInventory: [],
+  purchaseRequests: [],
+  inventoryUpdates: [],
+}
+
+interface inventoryDataSliceState {
+  products: Record<string, ProductRecord>;
+  sitesInventory: Record<string, SiteInventoryData>;
```
julianrkung: It's unclear what `string` refers to here as a key. I would probably try a soft type alias to make it easier on the devs reading later.

Something like 
```
type siteId = string;
type productId = string; 

interface inventoryDataSliceState {
  products: Record<productId, ProductRecord>;
  sitesInventory: Record<siteId, SiteInventoryData>;
}
```

src/lib/redux/siteDataSlice.ts
```diff
@@ -23,6 +23,7 @@ export const EMPTY_SITE: SiteRecord = {
   financialSummaryIds: [],
   financialSummaries: [],
   tariffPlans: [],
+  inventoryIds: [],
```
julianrkung: Sorry I forgot if we needed this or not. When will this be used?

src/screens/Inventory/AddInventory.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const inventory = JSON.parse(JSON.stringify(EMPTY_INVENTORY));
+    inventory.productId = selectedProductId;
+    inventory.siteId = currentSite.id;
+    inventory.currentQuantity = startingAmount;
+    inventory.periodStartQuantity = startingAmount;
+    addInventoryToRedux(inventory);
+    createInventory(inventory);
+
+    // TODO create inventory update
+    history.replace(`item`, { inventoryItem: inventory });
```
julianrkung, line 66: I know you probably took this idea from the `customerProfile`, but we're going to move to having a `currentCustomer` in redux and displaying the `currentCustomer` instead of passing state. We're doing this because the `Edit Customer` workflow is reached via the `BaseHeader` component, and there's not a good way to pass `customer` state to `BaseHeader`. 

The inventory should follow the same paradigm so everything is consistent. You can make a `setCurrentInventoryItem` reducer to set the inventory item appropriate. You should also make sure that when a `Site` context is changed via `setCurrSite`, we should null out the `currentInventoryItem`. This can be done via passing an `extraReducer` to `siteDataSlice` according to [these docs](https://redux-toolkit.js.org/api/createSlice#the-extrareducers-map-object-notation). I'll also have a draft PR posted of this for `customerDataSlice` by 3/22 if you want to wait for that

src/screens/Inventory/AddInventory.tsx
```diff
@@ -0,0 +1,99 @@
...
+    // Prevent page refresh on submit
+    event.preventDefault();
+
+    // Make a deep copy of an empty inventory record
+    const inventory = JSON.parse(JSON.stringify(EMPTY_INVENTORY));
+    inventory.productId = selectedProductId;
+    inventory.siteId = currentSite.id;
+    inventory.currentQuantity = startingAmount;
+    inventory.periodStartQuantity = startingAmount;
+    addInventoryToRedux(inventory);
```
julianrkung: The inventory's id should be added if the user is online

image

This can probably be done as a promise to `createInventory`, but be aware of the race condition between persisting the `inventory` object to redux and the API request finishing (since these 2 operations are happening in parallel right now). Might be good to just primitively `await` the API call in a try statement, grab the id (and catch if offline), and `addInventoryToRedux(inventory)` after placing the id in `inventory`

src/screens/Inventory/InventoryMain.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+function InventoryMain (props: InventoryProps) {
+    const { classes, sitesInventory, currentSite } = props;
+    const currentSiteInventory = sitesInventory[currentSite.id];
+    return (
+      <BaseScreen title="Inventory">
+        <BaseScrollView>
+          {currentSiteInventory.siteInventory.map((inventory: InventoryRecord) =>  (
+            <Link key={inventory.id} to={{ pathname: `/inventory/item`, state: { inventoryItem: inventory }}}>
```
julianrkung, line 36: Right now, key is dubious since ids are not unique (all offline will be empty string). Let's do the negative `id` for offline items? The more I think about different cases the more I like the idea

julianrkung (commented):
Looks really good so far! I mentioned some tiny gotchas and a change in paradigm (moving towards a `currentInventoryItem` etc). Looks like we can merge pretty soon though :)

julianrkung (commented):
Also, according to NPO, Products are site-specific (that is, when I am eventually able to create inventory types, the inventory type that I create should only be present in the site I created it in). I think this might not work with the current `Products` data because it assumes the same products are available to all sites

src/lib/redux/siteDataSlice.ts
wangannie: Good catch! Making it optional + removing it here

src/screens/Inventory/AddInventory.tsx
wangannie: Tabling this change to address in a dedicated PR for inventory profile 👍

src/lib/airtable/interface.ts
```diff
@@ -34,6 +37,11 @@ export interface SiteRecord {
   financialSummaryIds: string[];
   financialSummaries: FinancialSummaryRecord[];
   tariffPlans: TariffPlanRecord[];
+  inventoryIds?: string[];
```
julianrkung, line 41: What do you think about putting a comment saying that these are put into `InventoryDataSlice` and deleted from `SiteRecord`? So devs don't go searching?

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const EMPTY_SITE_INVENTORY_DATA : SiteInventoryData = {
+  siteInventory: [],
+  purchaseRequests: [],
+  inventoryUpdates: [],
+}
+
+// Aliases for readability
+export type siteIdString = string;
+export type productIdString = string;
```
julianrkung: I think types are capitalized. `SiteIdString` etc. In my PR for `customerData` I put these in `Interface.ts` but i'm also okay with them here (and I'll change them on my PR) unless you think. `Interface.ts` is a better spot

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -0,0 +1,73 @@
...
+const inventoryDataSlice = createSlice({
+  name: 'inventoryData',
+  initialState,
+  reducers: {
+    saveInventoryData(state, action) {
+      const {siteId, products, inventory, purchaseRequests, inventoryUpdates} = action.payload;
+      products.map((product: ProductRecord) => 
+        state.products[product.id] = product
+      );
+      const siteData = JSON.parse(JSON.stringify(EMPTY_SITE_INVENTORY_DATA));
```
julianrkung: NIT: might prefer a better name here since `siteData` alludes to information related to `siteDataSlice` imo, but the code is self-explanatory enough so not a blocker if you can't figure out a better name

src/screens/Inventory/AddInventory.tsx
```diff
@@ -0,0 +1,111 @@
...
+  userId: string;
+  sitesInventory: Record<siteIdString, SiteInventoryData>;
+}
+
+
+function AddInventory(props: AddInventoryProps) {
+  const { classes, currentSite, products, userId, sitesInventory } = props;
+  const history = useHistory();
+  // Products that the site already has inventory for
+  const currentSiteProducts = sitesInventory[currentSite.id].siteInventory.map((inventory: InventoryRecord) => inventory.productId);
```
julianrkung: Let's make the code more defensive. I think this errors if your `sitesInventory` returns the empty object you default it to. There might be other error cases as well

src/screens/Inventory/AddInventory.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </form>
+    </BaseScreen>
+  );
+}
+
+const mapStateToProps = (state: RootState) => ({
+  currentSite: state.siteData.currentSite || EMPTY_SITE,
+  products: state.inventoryData.products || {},
+  userId: state.userData.user?.id || '',
+  sitesInventory: state.inventoryData.sitesInventory || {},
```
julianrkung: Is there any reason we don't default to `EMPTY_SITE_INVENTORY_DATA` here and instead do an empty object?

src/screens/Inventory/AddInventory.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  return (
+    <BaseScreen title="New Inventory" leftIcon="backNav">
+      <form noValidate className={classes.content} onSubmit={() => false}>
+        <FormControl variant="outlined" className={classes.formControl}>
+          <Autocomplete
+            id = "select-item"
+            options = {productOptions}
+            style = {{ marginBottom: 24 }}
+            getOptionLabel={(productId) => products[productId]?.name + ' (' + products[productId]?.unit + ')'}
```
julianrkung: string templates (backtick strings) might be cleaner here. `${products[productId]?.name} (${products[productId]?.unit})` or whatever so you don't have to read the string concatenation with parentheses

src/screens/Inventory/AddInventory.tsx
```diff
@@ -0,0 +1,111 @@
...
+    inventory.periodStartQuantity = startingAmount;
+    
+    let inventoryId = "";
+    // Grab new inventory id from API response or generate temporary ID if offline
+    try {
+      await createInventory(inventory).then(resp => resp?.json()).then(data => inventoryId = data.id);
+    } catch (err) {
+      inventoryId = generateOfflineInventoryId();
+      console.error("[AddInventory] (handleSubmit) ", err);
+    }
```
julianrkung: This is my fault (I set a bad precedent with `addCustomer` etc..) but I don't think this try catch does what you want it to do (though it does work). 

Here's the error:
image

Since the error is already caught in the function in `request.js`, the function `createInventory` just returns null and we error again doing `data.id`. We could remove error handling on this level and put it all in `request.js` level (also moving `addInventoryToRedux` to that level as well probably like mentioned in [comments](https://github.com/calblueprint/meepanyar/pull/55/files?file-filters%5B%5D=.tsx#r598162072) in tiffany's PR) OR we could remove it on the request level and do it on this level. I probably prefer the former.

Either way, we could also save this for another PR and do it all at once later in the week (since this needs to be done for `addCustomer` as well

julianrkung (commented):
We're probably good to merge after this round of changes! Gave some comments

src/screens/Inventory/AddInventory.tsx
wangannie: Because sitesInventory is the map, but `EMPTY_SITE_INVENTORY_DATA` only represents the value

src/screens/Inventory/AddInventory.tsx
wangannie: I tried this but it was giving me a lot of trouble so I stuck to this. I'll look into it again!

src/screens/Inventory/AddInventory.tsx
wangannie: Went with the former and moved addInventoryToRedux into `createInventory`

src/screens/Inventory/AddInventory.tsx
julianrkung: Ah you're right, sorry I think I was confused for a moment, that's fine imo

wangannie: Re-formatted so it pulls site inventory data directly instead of the whole map!

src/lib/airtable/request.js
```diff
@@ -12,17 +12,23 @@
...
   createRecord,
   createRecords,
-  updateRecord,
-  updateRecords,
-  getAllRecords,
-  getRecordById,
-  deleteRecord,
+
+
+
```
julianrkung: Formatting issue came back

julianrkung (approved):
LGTM after formatting fix. Remember to put the patches in this PR before you merge (or create a new PR right after with the new patches if you'd want them separately reviewed)

PR #55, opened 2021-03-20
Tiffany/edit customer

cnnmon:

# What's new in this PR

You can use the [edit customer page](http://localhost:3000/customers/customer/edit) to edit customers in Airtable and Redux & create customer records!

## How to review

Edit customer -> check airtable
http://localhost:3000/customers/customer/edit

## Next steps

When you press the back button after editing, you go back to the old version of someone's profile (the changes are still made though, so idk whether that's important). Also weird errors in node console after you edit but it still works?

### Screenshots

CC: @julianrkung


src/screens/Customers/EditCustomer.tsx
```diff
@@ -15,28 +25,106 @@ const styles = (theme: Theme) =>
...
+  // TODO: Add form input validation and error messaging
+  const handleSubmit = (event: React.MouseEvent) => {
+    // Prevent page refresh on submit
+    event.preventDefault();
+
+    // Add other info necessary to create the Airtable record
+    let customerUpdate: CustomerUpdateRecord = {
+      id: '',
+      dateUpdated: formatUTCDateStringToLocal((new Date()).toString()),
+      customerIds: [currentCustomer.id],
```
julianrkung: This should probably just be `customerId` since there's at most 1 customerId being changed for a customer update. We can change that in a separate PR but let's put a TODO?

src/screens/Customers/EditCustomer.tsx
```diff
@@ -15,28 +25,106 @@ const styles = (theme: Theme) =>
...
+    customer.name = customerName;
+    customer.hasmeter = hasMeter;
+    customer.isactive = !customerInactive;
+    customer.tariffPlanId = selectedTariffPlanId;
+
+    if (customer.customerUpdates === undefined) {
+      customer.customerUpdates = [customerUpdate];
+    } else {
+      customer.customerUpdates.push(customerUpdate);
+    }
```
julianrkung: We probably don't actually need this since it's never displayed in the app (mostly just something for the admin to see via Airtable). We might be able to just create the customer update on the backend (and not create one on the frontend)

We could probably delete this from the interface also for now if that's an issue

src/lib/airtable/interface.ts
```diff
@@ -65,9 +65,9 @@ export interface CustomerRecord {
 export interface CustomerUpdateRecord {
   id: string;
   dateUpdated: string;
-  customerId: string[];
+  customerIds: string[];
```
julianrkung: I think we should make this a string since there's only 1 customer (might need to change some stuff in Airtable if it's foreign-key-many but that can be separate pr)

src/lib/redux/siteDataSlice.ts
```diff
@@ -1,17 +1,19 @@
 /* eslint-disable no-unused-vars */
 /* eslint-disable no-param-reassign */
 import { createSlice } from '@reduxjs/toolkit';
-import { SiteRecord, FinancialSummaryRecord } from '../airtable/interface';
+import { SiteRecord, CustomerRecord, FinancialSummaryRecord } from '../airtable/interface';
 
 interface siteDataSliceState {
   isLoading: boolean;
   currentSite: any;
+  currentCustomer: any;
```
julianrkung: Should this be a `customerRecord`?

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -54,6 +54,8 @@ function CustomerProfile(props: CustomerProps) {
   const { classes, match, currentSite } = props;
   const customer: CustomerRecord = props.location.state.customer;
 
+  setCurrentCustomer(customer);
```
julianrkung: We should be consistent in either passing the `customer` via `Link` or using `currentCustomer` in redux. We don't have to do it in this PR, if you want we can put a TODO (and tag me so I do it)

src/lib/redux/siteDataSlice.ts
```diff
@@ -69,9 +92,15 @@ const siteDataSlice = createSlice({
         ]
         }
       }
-      }
+    },
+    editCustomer(state, action) {
+      const stateCopy = JSON.parse(JSON.stringify(state));
+      const index = stateCopy.currentSite.customers.findIndex((e: CustomerRecord) => e.id === action.payload.id);
+      stateCopy.currentSite.customers[index] = action.payload;
```
julianrkung: If the customer we are editing is the same as the current customer (which might always be the case?), we need to update the `currentCustomer`'s object so that it also reflects the changes.

julianrkung: Maybe the ^ should happen in `editCustomerInRedux`? Do not know, whatever you think is cleaner

src/lib/redux/siteDataSlice.ts
```diff
@@ -69,9 +92,15 @@ const siteDataSlice = createSlice({
         ]
         }
       }
-      }
+    },
+    editCustomer(state, action) {
+      const stateCopy = JSON.parse(JSON.stringify(state));
+      const index = stateCopy.currentSite.customers.findIndex((e: CustomerRecord) => e.id === action.payload.id);
```
julianrkung: I think this might miss if the customer was created offline --> edited offline. Let's talk about how that should work offline, because we can handle it the same way we handle payments and meter readings

julianrkung (commented):
Looks really good as a first iteration, put some comments down!

src/screens/Customers/EditCustomer.tsx
```diff
@@ -15,28 +25,106 @@ const styles = (theme: Theme) =>
...
+
+    // TODO: add error handling
+    editCustomerInRedux(customer);
```
julianrkung: Actually, this `editCustomerInRedux` call might even be better placed inside of `editCustomer`. Then, we can abstract away the `editCustomer` function as BOTH a frontend change and a backend change and ensure they're consistent since there should never be a case where I update the frontend and not the backend and vice verase. 

This applies to `addCustomerToRedux` as well. Thoughts on this? @wangannie

src/screens/Customers/EditCustomer.tsx
cnnmon: I don't need to delete customerUpdates from interface for functionality because it's taking from the currentCustomer's data, but let me know if you want me to delete it to clean things up in the future!

src/lib/redux/siteDataSlice.ts
cnnmon: Should currentSite be SiteRecord then?

src/lib/airtable/interface.ts
cnnmon: I changed schemas to foreign-key-one to make sure everything worked here! What else do I need to do past that?

src/components/TextField.tsx
```diff
@@ -35,6 +36,7 @@ function Field(props: TextFieldProps) {
         InputProps={{ style: { fontSize: 14 }, disableUnderline: true, autoComplete: 'off' }}
         InputLabelProps={{ shrink: true }}
         color={props.primary ? 'primary' : undefined}
+        value={props.value ? props.value : undefined}
```
wangannie, line 39: Nit: empty string instead of undefined

src/lib/redux/siteDataSlice.ts
wangannie: yes^

src/screens/Customers/EditCustomer.tsx
```diff
@@ -15,28 +25,106 @@ const styles = (theme: Theme) =>
...
+    },
   });
 
 interface EditCustomerProps extends RouteComponentProps {
-  classes: { header: string; content: string };
+  classes: { header: string; content: string; formControl: string; };
   location: any;
+  currentCustomer: CustomerRecord;
+  currentSite: SiteRecord;
+  user: User;
```
wangannie, line 37: Not exactly related to this line/this probably isn't really relevant to your PR but it seems like `UserFields` is pretty out of date?

src/screens/Customers/EditCustomer.tsx
```diff
@@ -15,28 +25,106 @@ const styles = (theme: Theme) =>
...
 
 function EditCustomer(props: EditCustomerProps) {
-  const { classes } = props;
+  const { classes, currentCustomer, currentSite, user } = props;
+  const history = useHistory();
+
+  const [selectedTariffPlanId, setSelectedTariffPlanId] = useState(currentCustomer.tariffPlanId);
+  const [customerName, setCustomerName] = useState(currentCustomer.name);
+  const [reason, setReason] = useState("");
+  const [hasMeter, setHasMeter] = useState(currentCustomer.hasmeter);
```
wangannie, line 47: I see there's a `hasMeter` but not a way to edit the meter number -- if the customer used to have a meter number but was edited so that this is un-selected, it doesn't clear the meter number.

src/screens/Customers/EditCustomer.tsx
```diff
@@ -15,28 +25,106 @@ const styles = (theme: Theme) =>
...
+  const [customerInactive, setCustomerInactive] = useState(currentCustomer.isactive);
```
wangannie: You should negate the `currentCustomer.isactive` -- it's displaying the opposite every time

src/screens/Customers/EditCustomer.tsx
```diff
@@ -15,28 +25,106 @@ const styles = (theme: Theme) =>
...
+  const handleSubmit = (event: React.MouseEvent) => {
+    // Prevent page refresh on submit
+    event.preventDefault();
+
+    // Add other info necessary to create the Airtable record
+    let customerUpdate: CustomerUpdateRecord = {
+      id: '',
+      dateUpdated: formatUTCDateStringToLocal((new Date()).toString()),
+      customerIds: [currentCustomer.id],
+      explanation: reason,
```
wangannie: Since it's Explanation in Airtable, wouldn't it make sense to keep it consistent throughout? Or just change it to Reason in Airtable?

src/screens/Customers/EditCustomer.tsx
wangannie: Ooh yes agreed.

src/screens/Customers/EditCustomer.tsx
```diff
@@ -15,28 +25,106 @@ const styles = (theme: Theme) =>
...
+    setSelectedTariffPlanId(event.target.value as string);
+  }
+
+  // TODO: Add form input validation and error messaging
+  const handleSubmit = (event: React.MouseEvent) => {
+    // Prevent page refresh on submit
+    event.preventDefault();
+
+    // Add other info necessary to create the Airtable record
+    let customerUpdate: CustomerUpdateRecord = {
```
wangannie: Change to const

src/screens/Customers/CustomerProfile.tsx
wangannie: You can resolve the weird node error if you wrap this in a hook:

```
  useEffect(() => {
    setCurrentCustomer(customer);
  },[]); 
```

Source: https://stackoverflow.com/questions/62336340/cannot-update-a-component-while-rendering-a-different-component-warning

src/screens/Customers/EditCustomer.tsx
```diff
@@ -15,28 +25,106 @@ const styles = (theme: Theme) =>
...
+        <TextField label={'Reason'} id={'reason'} onChange={handleReasonInput}/>
+        <Button label={'SAVE'} onClick={handleSubmit} />
       </form>
     </BaseScreen>
   );
 }
 
-export default withStyles(styles)(EditCustomer);
+const mapStateToProps = (state: RootState) => ({
+  currentCustomer: state.siteData.currentCustomer || EMPTY_CUSTOMER,
```
wangannie, line 120: I'm confused why `currentCustomer` is part of `siteData` at all. Why is this necessary/why can't it just be passed around where needed? Also because it's not being reset/cleared when you  navigate away from the customer profile.

wangannie: Ok yeah I just saw Julian's note on this (https://github.com/calblueprint/meepanyar/pull/55/files#r598159199)

src/lib/redux/siteDataSlice.ts
```diff
@@ -25,6 +27,24 @@ export const EMPTY_SITE: SiteRecord = {
   tariffPlans: [],
 };
 
+export const EMPTY_CUSTOMER: CustomerRecord = {
```
wangannie: This is a duplicate! It already exists in `customerUtils`

src/screens/Customers/EditCustomer.tsx
wangannie: Actually relying on `currentCustomer` instead of props for this screen could resolve your issue of having to go back twice (I spent a while digging into react router trying to come up with a fix for that but I couldn't find much).

wangannie (commented):
Looking good!  Noted some cleanup stuff + some questions about the workflow. Overall, I'm really confused why currentCustomer is necessary/ whether it's a good idea to keep.

Also, in general, it would help if you added some more detail to your PR description! Specifically more detail on how to review (at least linking the backend PR) + giving more explicit steps of sample workflows (including edge cases) to test and results to expect. Detailed PR descriptions also serve as really good documentation to reference later on, which makes writing handoff documentation a lot easier!

Higher-level question: what's the point of making a customer update record? Besides adding the explanation, it isn't tracking anything, so I'm not sure what the utility is.

src/screens/Customers/EditCustomer.tsx
julianrkung: +1 should be consistent, thanks for catching

julianrkung:
> Higher-level question: what's the point of making a customer update record? Besides adding the explanation, it isn't tracking anything, so I'm not sure what the utility is.

Good question. It's solely for the NPO to be able to see via the Airtable (I don't believe there are plans to access it in the app, but it's a paper trail)

src/lib/redux/siteDataSlice.ts
```diff
@@ -69,9 +92,15 @@ const siteDataSlice = createSlice({
         ]
         }
       }
-      }
+    },
+    editCustomer(state, action) {
+      const stateCopy = JSON.parse(JSON.stringify(state));
```
julianrkung: We can write the code as if the redux store is mutable. This is my bad, sorry.

src/screens/Customers/EditCustomer.tsx
cnnmon: So to make sure, people will generally be moved to new meter numbers?

src/components/TextField.tsx
cnnmon: if i put it as an empty string it doesn't let anyone type anything

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -54,6 +54,10 @@ function CustomerProfile(props: CustomerProps) {
   const { classes, match, currentSite } = props;
   const customer: CustomerRecord = props.location.state.customer;
 
+  useEffect(() => {
+    setCurrentCustomer(customer); //TODO: @julianrkung need to make customer data retrieval consistent (Link vs Redux)
```
julianrkung, line 58: Note to other reviewers: I'll make sure to get rid of this before merging https://github.com/calblueprint/meepanyar/pull/56 . I plan on merging the `customerData` refactor PR AFTER this editCustomer PR is merged

src/lib/airtable/request.js
```diff
@@ -508,18 +509,22 @@ export const updateCustomer = async (id, recordUpdates) => {
...
-        'Content-Type': 'application/json'
-      },
-      body: JSON.stringify(customer)
-    })
-
-    console.log(resp);
-  } catch (err) {
-    console.log(err);
+  if (!customer.id) {
+    addToOfflineCustomer(customer, 'edits', customer.customerUpdate);
```
julianrkung: Note to other reviewers: This workflow is consistent with the offline workflow we have for payments and meter reading functionality. This'll probably need to be added to / adjusted in the future once we have more functionality set up, but I'm okay with approving this for now and dealing with it down the road.

src/screens/Customers/EditCustomer.tsx
```diff
@@ -15,28 +24,99 @@ const styles = (theme: Theme) =>
...
+  // TODO: Add form input validation and error messaging
+  const handleSubmit = (event: React.MouseEvent) => {
+    // Prevent page refresh on submit
+    event.preventDefault();
+
+    // Add other info necessary to create the Airtable record
+    const customerUpdate: CustomerUpdateRecord = {
+      id: '',
+      dateUpdated: formatUTCDateStringToLocal((new Date()).toString()),
+      customerId: currentCustomer.id, //TODO: may need to change Airtable from foreign-key-many
```
julianrkung: is this TODO already done?

src/lib/redux/siteDataSlice.ts
```diff
@@ -69,9 +74,14 @@ const siteDataSlice = createSlice({
         ]
         }
       }
-      }
+    },
+    editCustomer(state, action) {
+      const index = state.currentSite.customers.findIndex((e: CustomerRecord) => e.id === action.payload.id);
```
julianrkung, line 79: Can you put a note that this doesn't currently work for customers created while offline since they don't have an ID? We can address it in a future PR but we should definitely put that down so we remember.

src/lib/airtable/request.js
```diff
@@ -508,18 +509,22 @@ export const updateCustomer = async (id, recordUpdates) => {
 
 // NONGENERATED: Edit customer
 export const editCustomer = async (customer) => {
```
julianrkung: I'd be careful with the variable naming here. This object is a `customer` with an added `customerUpdate` in it. Therefore, when you do `editCustomerInRedux(customer)`, the edited customer you added now looks different than all the other customers (the edited customer has a `customerUpdate` field that no other customer has).

You could:
1. pass them as 2 separate parameters initially and when you do the fetch you can `JSON.stringify({...customer, customerUpdate})` like how you did when calling `editCustomer`)
2. If you look at the custom backend endpoint, [here](https://github.com/calblueprint/meepanyar-node/pull/15/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R125), we really just separate out the data and do a `updateCustomer` and `createCustomerUpdate`. I think we could instead just separate out the data here and call `updateCustomer` and `createCustomerUpdate` here and not need the custom backend endpoint. I'm not sure if I'm missing something with this solution though, so let me know if you have concerns. 

If solution 2 doesn't have problems, I'd go with that since the frontend has built-in error handling (thanks to the offline library we're using), so we'll retry in the case either our network goes down or the backend is down etc. We could also just go with solution #1 for now and revisit solution #2 later. Your call

julianrkung (commented):
It's looking really good, I really only had the comment in `editCustomer`. I think after this iteration, we should be good to merge!

src/lib/airtable/request.js
julianrkung: We also have to editCustomerInRedux after the if statement to update the redux store (maybe this just happens after the `if / else`?

julianrkung (approved):
Approved knowing of bug where you return to a previously un-edited profile. Will fix in a future PR. There's a branch where this is fixed: https://github.com/calblueprint/meepanyar/compare/tiffany%2Fedit-customer...annie%2Fedit-customer-suggestions?diff=split&expand=1 

that can fix the bug

PR #56, opened 2021-03-22
Add CustomerData slice to redux

julianrkung:

# What's new in this PR
This PR is a refactor of our redux store to add a `CustomerDataSlice`. This was done for the following reasons
1. To make the division of data in our redux store nicer (less need to nest deeply into `siteData` to grab properties)
2. Since Redux data is immutable, un-nesting our data is advantageous because we do less copying when we persist new data (such as a customer or meter reading) into the redux store
3. There was some trouble with the data flow in getting the customer in `CustomerProfile` to the `Edit Customer` screen. We remedy this by adding a global `currentCustomerId` to the redux store, which will be the customer that will be edited or displayed when navigating to the `Edit Customer` or `CustomerProfile` screens respectively.
    - **Note**: We should standardize the ^ usage if we adopt it. That is, we should no longer pass state via Link for things like Customer, and we should be consistent with always grabbing the `currentCustomer` via the redux store.

TODO BEFORE MERGE: Recreate patch for request.js

## Changelog
- Create a `customerData` slice in Redux with a `currentCustomer` attribute
- Created getter and setter methods for customerData
- `/customers/customer` now displays whatever the customer corresponding to`customerData.currentCustomerId` from redux. `Link` and `addCustomer` functionality was changed to set this value before navigating to `CustomerProfile`
- No functional changes should be observed (only state management)
- Persist IDs from Airtable for Customers who've just been created
- Airtable customers made while the app was offline will receive a temporary offline ID
- Changed `EditCustomer` to `goBack` instead of replace

**The backend branch `julian/create-customer-id` MUST BE USED for this PR**
PR for backend: https://github.com/calblueprint/meepanyar-node/pull/18

## How to review
- Make sure you've checked out the correct backend branch from the PR [here](https://github.com/calblueprint/meepanyar-node/pull/18)
- Log in and check the redux store's `customerData` is populating correctly.
- Navigate to all customer-related links and make sure they work properly
- Create a new customer and ensure sure that flow still works properly
  - You are taken to the new customer's profile
  - Going back to the customer list, you can find the customer that was just created
  - Attempt to "Edit" the customer you just created. After editing, ensure that the edits are accurately reflected on `customerMain`, `customerProfile` and on the airtable base.

### Online sources
https://redux-toolkit.js.org/api/createEntityAdapter
https://redux-toolkit.js.org/api/createSelector

### Related PRs

## Next steps
- Refactor `siteData` and add useful `getter` functions for that
- Make offline ID generate robust
- Persist IDs from Airtable for Customers who've just been created
- Debatable: Refactor `customerData` to a `Record<SiteId, Record<CustomerId, CustomerRecord>>` once offline IDs are added
- Use Selectors for `getter` functions like `getCurrentCustomer` and `getAllCustomersInSite`
- Potentially switch to [createEntityAdapter](https://redux-toolkit.js.org/api/createEntityAdapter) in the future

### Screenshots

CC: @julianrkung


src/screens/Customers/CustomerProfile.tsx
```diff
@@ -162,6 +163,7 @@ function CustomerProfile(props: CustomerProps) {
 
 const mapStateToProps = (state: RootState) => ({
   currentSite: state.siteData.currentSite || EMPTY_SITE,
+  customer: getCurrentCustomer(),
```
julianrkung: I'm also down to just do `state.customerData.currentCustomer` and removing the getter method entirely. I initially wanted to do the `getter` in the case that our `state.customerData.siteToCustomers` was in the form of `Record<SiteId, Record<CustomerId, CustomerRecord>>` instead of `Record<SiteId, CustomerRecord[]>`

src/lib/redux/siteData.ts
```diff
@@ -1,8 +1,11 @@
 import { getAllSites } from '../airtable/request';
-import { addCustomer, saveSiteData, setCurrSite, setLoadingForSiteData } from './siteDataSlice';
+import { SiteRecord, CustomerRecord, SiteId } from '../airtable/interface';
+import { EMPTY_SITE, saveSiteData, setCurrSite, setLoadingForSiteData } from './siteDataSlice';
+import { saveCustomerData } from './customerDataSlice';
 import { store } from './store';
 
-const refreshSiteData = async (loadSilently: boolean): Promise<void> => {
+//  TODO: DON'T KNOW WHERE TO PUT THIS????? Shouldn't be here.
```
julianrkung: I was thinking about moving this function to `userData`, but otherwise couldn't find a proper place for it.

src/lib/redux/customerDataSlice.ts
```diff
@@ -0,0 +1,69 @@
...
+    paymentIds: [],
+    payments: [],
+    customerUpdateIds: [],
+    customerUpdates: [],
+    totalAmountBilledfromInvoices: 0,
+    totalAmountPaidfromPayments: 0,
+}
+
+// TODO: @julianrkung, We should eventually make this a Record<SiteId, Record<CustomerId, CustomerRecord>> map for efficiency
+// We can't do this yet because customers created while offline do not receive an id yet
```
annieyro, line 26: Non-blocking, but is this actually solvable (aren't the IDs generated by Airtable - hence the create call needs to go through to get the customer's ID at all?)

src/lib/redux/siteData.ts
annieyro: It doesn't reallyy matter, but I feel like I might just create a new file with literally one function if you want to make sure logic of where things go is consistent 😅  - maybe call it `async`?

src/lib/redux/siteData.ts
```diff
@@ -1,8 +1,11 @@
...
+const refreshData = async (loadSilently: boolean): Promise<void> => {
 
   //if background is true then dont set loading parameter.
```
annieyro: nit: this comment seems out of date

src/lib/redux/siteDataSlice.ts
```diff
@@ -9,6 +9,7 @@ interface siteDataSliceState {
   sites: any[];
 }
 
+//TODO @julianrkung: Change sites to siteIdsToSites and change currentSite to currentSiteId
```
annieyro, line 13: nit: lacking context, but can I confirm this is a note to refactor s.t the `sites` object is now a map, not an array? (i.e not just renaming)

src/screens/Customers/CustomerProfile.tsx
annieyro: Yeah, I think that makes more sense - it's better to be consistent (so match the above line if possible). You can always refactor later if necessary

src/lib/redux/siteData.ts
```diff
@@ -19,19 +22,32 @@ const refreshSiteData = async (loadSilently: boolean): Promise<void> => {
...
+  store.dispatch(saveCustomerData({ siteIdsToCustomers }));
   store.dispatch(saveSiteData(siteData));
 };
 
+const extractCustomerDataFromSite = (sites: SiteRecord[]): Record<SiteId, CustomerRecord[]> => {
+
+  const siteIdsToCustomers: Record<SiteId, CustomerRecord[]> = {};
+  sites.map((site: SiteRecord) => {
+    siteIdsToCustomers[site.id] = site.customers || [];
+    delete site.customers;
```
annieyro: nit: add a comment here - deleting makes me nervous in general lol, but AFAICT when necessary you need to use `getAllCustomersInSite` instead for the main immutability issue etc. I think it'd be useful to add a note inline

annieyro (commented):
LGTM from code quality standpoint (didn't test any functionalities) with so many moving pieces (!).

General comment - it would be helpful if you (and by extension the rest of the team) added more context for the problem/solution being addressed in the PR. I think having to articulate your motivation and reasoning for why you did what you did also helps to catch logical errors etc, but it's useful for knowledge sharing and for people catching up async

src/lib/redux/customerDataSlice.ts
julianrkung: We're deciding what to do here. I was tossing around the idea of having temporary negative IDs for customers created offline (once the customer is created, we'll pull from airtable and override with the real ID). I think there's a need to be able to uniquely identify customers created even while offline, and I don't know a better solution.

src/screens/Customers/CustomerProfile.tsx
julianrkung: Ok yeah, good point. I'll change this to be consistent and get rid of `getCurrentCustomer` for now since this was its only use.

src/lib/redux/siteData.ts
julianrkung: That's a good point. I'll probably just create the new file.

As for async: I was afraid `async` in the name might be misleading because our first login call makes it kinda synchronous to the user (the first `refreshData` passes `loadSilently=false`) so that they can't access the rest of the data until the pull succeeds. Not entirely sure what's standard though

src/lib/redux/siteDataSlice.ts
julianrkung: Yes, this is a note to refactor `sites` into an object `siteIdsToSites: Record<SiteId, SiteRecord>` and have `currentSite` become a string `currentSiteId` that will be used to index into `siteIdsToSites`. This is done mostly because we're shifting to using `id` as lookups.

The refactor will also put a wrapper over the `Site` portion of redux store so it's easier to interface with, so adding stuff like `getCurrentSite`, etc...

src/lib/utils/offlineUtils.js
```diff
@@ -72,3 +72,10 @@ export const addToOfflineCustomer = async (customer, fieldToAppend, objectToAdd)
         }
     }
 }
+
+// TODO: Return with robust ID generation. 
+// We could probably have all offlineIDs be generate through 1 function if we pass
+// the redux slice / entity to generate the ID for?
+export const generateOfflineId = () => {
+    return (Math.floor(Math.random() * 1000)).toString();
+}
```
julianrkung, line 81: I was thinking we could jsut make a generic `generateOfflineId` instead of 1 for each object type (Inventory, Customer, CustomerPurchase, etc...) since they'd presumnably be generated via the same routine. I'm not sure what the logic would look like (they might just be case-by-case logic so it might be kinda ugly). lmk what you think or if you prefer to separate them out, I'm good with either approach.

cc @wangannie

cnnmon (approved):
everything looks good to me! good job

src/lib/airtable/interface.ts
```diff
@@ -33,7 +33,7 @@ export interface SiteRecord {
   id: string;
   name: string;
   customerIds: string[];
-  customers: CustomerRecord[];
+  customers?: CustomerRecord[];
```
wangannie: Nit: later when you  merge `main` in, move this down to the section of other optional things (inventory, products, etc) that are just sent along with the site record

src/lib/redux/customerDataSlice.ts
```diff
@@ -0,0 +1,79 @@
...
+    currentCustomerId: EMPTY_CUSTOMER.id,
+};
+
+const customerDataSlice = createSlice({
+    name: 'customerData',
+    initialState,
+    reducers: {
+        saveCustomerData(state, action) {
+            const { siteIdsToCustomers } = action.payload;
+            state.siteIdsToCustomers = siteIdsToCustomers;
```
wangannie: I'm pretty sure you can directly set `site.siteIdsToCustomers = action.payload` if you take out the `{}` wrapping where this is called, though you might have had other motivations for doing this?

src/lib/utils/offlineUtils.js
wangannie: Sounds good. Just wondering, how does this relate to what we were talking about wrt tracking offline purchase requests using timestamps? Or would we use generated offline ids regardless and just use timestamps to match them?

src/lib/airtable/request.js
```diff
@@ -89,11 +91,16 @@ export const createCustomer = async (customer) => {
...
-
     console.log(resp);
+    await resp.json().then(data => customerId = data.id);
   } catch (err) {
     console.log(err);
+    customerId = generateOfflineId();
   }
+  customer.id = customerId;
+  addCustomerToRedux(customer);
+  setCurrentCustomerInRedux(customer);
```
wangannie: Since this isn't directly related to creating the customer, I think it might make more sense to move this to the actual `handleSubmit` right before navigating?

src/screens/Customers/CustomerMain.tsx
```diff
@@ -218,14 +220,14 @@ function CustomerMain(props: RouteComponentProps & UserProps) {
       <BaseScrollView>
         <FormHelperText>{filterLabels[0]}</FormHelperText>
         {filteredCustomers.map((customer, index) => (
-          <Link key={index} to={{ pathname: `${props.match.url}/customer`, state: { customer: customer } }}>
+          <Link key={index} to={{ pathname: `${props.match.url}/customer` }} onClick={() => setCurrentCustomerInRedux(customer)} >
```
wangannie: Nit: you could change this back to just `to={path}` again. Same on line 230

src/lib/utils/customerUtils.ts
```diff
@@ -3,9 +3,9 @@ import { CustomerRecord, MeterReadingRecord, SiteRecord, TariffPlanRecord } from
 import { getCurrentPeriod, getDiffinPeriods, getPeriodFromDate } from '../../lib/moment/momentUtils';
 
 const isCurrentReading = (mr: MeterReadingRecord): boolean => {
-    const period: string = getPeriodFromDate(mr.date);
-    const currPer: string = getCurrentPeriod();
-    return period === currPer;
+  const period: string = getPeriodFromDate(mr.date);
+  const currPer: string = getCurrentPeriod();
+  return period === currPer;
```
wangannie: Nit: this feels like a lot of unnecessary shortening of variable names that doesn't help with readability

wangannie (approved):
This looks really good! I just left a few comments on some small details. I didn't look too closely into the getters/setters since I think a lot of that can be cleaned up once you implement `createEntityAdapter` and selectors!

src/lib/utils/customerUtils.ts
julianrkung: i agree. I'm reverting this change to preserve blame (my linter just indented here)

src/lib/airtable/request.js
julianrkung: Good catch, I agree. Thanks!

src/lib/redux/customerDataSlice.ts
julianrkung: I don't think I had a solid reason. I agree, looks kinda stupid. Changed!

src/lib/utils/offlineUtils.js
julianrkung: I was thinking we use generated offline IDs regardless and just use timestamps to match. We'll probably find some problems with whatever method we choose during implementation and we can discuss then?

PR #57, opened 2021-03-22
[Translation Support] Translate between English and Burmese (Unicode + Zawgyi) 

micahtyong:

# What's new in this PR

In this PR, I address [step 2](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#13da3cf317c04ee18b0c610d0b6415ff) and [step 3](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#15b596e89b9043ddbb92ae1911beff26) of our [roadmap for translation support](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7) by creating a custom React hook that handles internationalization in the Mee Panyar app. 
- `lib/i18next/resources` and `lib/i18next/encodings` contain rules for translating between English and Burmese, and Unicode to Zawgyi, respectively. Thus, their diffs will be large. 

## How to review
1. The new React hook is in `src/lib/i18next/translator.ts`. Import it into a file of your choosing (see example below). 
2. Initialize the hook with `const intl = useInternational();`. 
3. Wrap a hard-coded string with `{intl(<hard coded string>)}`, then upon rendering, you should see the results displayed in Burmese. 

**NOTE:** The translations are contained in `lib/i18next/resources.js` but seem to only be representative of text in the Adalo app. Thus, you should choose a hard-coded text that is currently in `lib/i18next/resources.js`; otherwise, this will not work. 

## Relevant Links

### Related PRs
- [Fetching and parsing translations](https://github.com/calblueprint/meepanyar-node/pull/10)

## Next steps

- [Automatic language configuration](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#222323ac6b0b4ada80b8c801898abf1d)
- Propagating this hook throughout the entire app (essentially comes down to replacing any hardcoded strings with our new `useIntl` hook). 
- Updating the translations Airtable as necessary (currently only supports text in the Adalo app) 
- Regenerate `resources.js` files whenever doing production builds

### Screenshots

For example, in `CustomerMain.tsx`, we can import our hook with
```
import { useIntl } from '../../lib/i18next/translator'; 
```
then use it in `CustomerMain` with 
```
const intl = useIntl(); 
...
<h1 className={classes.title}>{intl("Customers")}</h1>
...
```

If the user's default language setting is English, they will get the following result: 
Screen Shot 2021-03-22 at 12 10 46 AM

If the user's default language setting is Burmese and their encoding is in Unicode, they will get the following result: 
Screen Shot 2021-03-22 at 12 05 57 AM

If the user's default language setting is Burmese and their encoding is in Zawgyi, they will get the following result: 
Screen Shot 2021-03-22 at 12 05 27 AM

CC: @julianrkung


src/lib/i18next/translator.ts
```diff
@@ -0,0 +1,27 @@
...
+enum LanguageEncoding { 
+    UNICODE,
+    ZAWGYI 
+}
+
+/* Global variable that determines user encoding */
+// TODO Step 4: Interpret encoding based on user settings
+let userEncoding: LanguageEncoding = LanguageEncoding.UNICODE; 
+
+export const useIntl = () => {
```
julianrkung: NIT (non-blocker): I was trying to think of a better hook name because I think `useIntl` isn't very expressive for devs trying to pick this up (since they likely won't know what intl stands for). I couldn't think of a good alternative, but if you can think of anything more descriptive before you merge, that'd be great.

julianrkung (approved):
Great work! :) LGTM! We might want to autogenerate the `src/lib/i18next/resources.js` in the future, since the NPO will be interfacing with Airtable to make changes. I think that's already on your roadmap though (and there's the PR in mp-node that already might do this)

src/lib/i18next/translator.ts
micahtyong: Gotcha. I'll change it to `useInternationalization` instead, as that seems more clear while keeping things concise. If we're _really_ going for clarity, I can change it to `useLanguageInternationalization`, but that's a pretty long name for a hook.

PR #58, opened 2021-03-23
Update Workbox V6 and Sevice worker caches all endpoints for offline

julianrkung:

# What's new in this PR
Previously, the service worker would only launch on the `/` endpoint. This can be shown on the `main` branch, where production builds will respond with `offline` page error when refreshing on anything other than the `/` endpoint. 

This PR expands the service worker URL scope to all frontend endpoints, so refreshes on endpoints other than the `/` when offline will still work. This PR also updates to Workbox v6. We were previously on workbox V3.

## How to review
1. Run `npm run build` to build a production version of the app
2. Run `serve -s build` to run the production version of the app on localhost
3. Log into the app on localhost using the production site
4. Shut down the frontend server running on the terminal
5. Do a refresh of the app in browser and see that it still correctly loads
6. Open a new tab and type in an endpoint in the app and see it still correctly loads even with the frontend offline (I went to `http://localhost:5000/customers`)

### Online sources
Migration from V3 --> V6 Changelog
[Plugin name change](https://developers.google.com/web/tools/workbox/guides/migrations/migrate-from-v4#plugin_classes_renamed)
[Route caching navigation](https://developers.google.com/web/tools/workbox/guides/migrations/migrate-from-v4)
[Strategies constructor](https://developers.google.com/web/tools/workbox/guides/migrations/migrate-from-v3)

### Related PRs

## Next steps

### Screenshots

CC: @julianrkung

### DIFF
--- a/src/custom-sw/sw-custom.js
+++ b/src/custom-sw/sw-custom.js
@@ -26,6 +26,9 @@ if ("function" === typeof importScripts) {
         // all assets under build/ and 5MB sizes are precached.
         workbox.precaching.precacheAndRoute([]);
 
+        // React is an SPA, so all routes we define are cached and mapped into index.html
+        workbox.routing.registerNavigationRoute('/index.html')
+
         // Font caching
         workbox.routing.registerRoute(
             new RegExp("https://fonts.(?:.googlepis|gstatic).com/(.*)"),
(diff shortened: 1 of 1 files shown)


micahtyong (approved):
Nice, this looks good to me. I was able to go through the review successfully and tested on `http://localhost:5000/incidents`. For future reviewers, be sure to fetch and pull from master on both `meepanyar` and `meepanyar-node`. Also, and I doubt anyone else will run into this, but I already had a global Python command called `serve`, so I had to call it using the [absolute path to the npm version](https://stackoverflow.com/questions/17639831/what-happens-if-two-command-line-programs-share-the-same-name).

PR #60, opened 2021-03-24
[Inventory] Create and review purchase requests + normalize inventoryData

wangannie:

# What's new in this PR
This PR allows users to:
- Submit purchase requests for specific inventory items
- Approve or deny purchase requests (currently any user can do this, not limited to admins)
Note that this only covers bare functionality. I'll continue to update the front-end as the designs evolve.

**Additional restructuring & cleanup**
Made the following updates inspired by Redux docs ([normalizing state shape](https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape), [createEntityAdapter](https://redux.js.org/tutorials/fundamentals/part-8-modern-redux#using-createentityadapter)):
- Switched from using `mapStateToProps` to  `useSelector`
- Refactored `inventoryData` to normalize using `createEntityAdapter`for inventory updates, products, site inventory, and purchase requests.

**TODO before merging:**

- [x] Update patch for `requests.js`

## How to review
On the backend, run the latest version of `main`.

**Create a purchase request**
1. In the Inventory tab, select an item and click **Purchase**
2. Fill out the form (note: you can't add a receipt yet) and hit **Confirm**
3. This should take you back to the inventory profile. Verify that:
    - The inventory quantity is updated (old qty + purchase request qty)
    - The request was correctly created in Airtable
    - The request was created in Redux with a valid record id if online, or a randomly generated integer if offline. Note: if you're getting an integer id when online, check the console for an error message.
4. Navigate back to the main Inventory screen and select **Purchase Requests** at the top of the list (this is a placeholder for navigation, pending designs). You should see the purchase request you just created at the top of the **Pending Review** section.

**Review the purchase request**
1. In the **Purchase Requests** screen, select a purchase request under **Pending Review** and click either **Approve** or **Deny**.
2. This should take you back to the list of purchase requests where the request you just reviewed will move from the **Pending** section to either the **Approved** or **Denied** section (scroll down).
3. Verify that the update is reflected in Airtable and Redux (`status`, `reviewerId`, and `approvedAt` should all change).

Test this with both Approve and Deny. If you approve the request, it should update the inventory item's current quantity as well (check Airtable and Redux).

**Test redirects**
1. Reset the Redux store
2. Go directly to http://localhost:3000/inventory/purchase-requests/purchase-request by pasting it into the URL bar
3. Since there is no `currentInventoryId` set in state, it should redirect you to the main inventory screen.

**Switching between sites**
1. In the East Asian Library site, navigate to Inventory and select a product (e.g. Diesel) to view the profile page. In the Redux inspector, `currentInventoryId` under `inventoryData` should match the record id of the selected item.
2. Go back to the main page `/home` and switch to a different site. In the Redux inspector, `currentInventoryId` should be reset to an empty string.

**Edge cases**
- After going through this workflow for a site, try changing the site and making sure it still works.

### Online sources
https://react-redux.js.org/using-react-redux/connect-mapstate
https://momentjs.com/docs/#/utilities/
https://www.typescriptlang.org/docs/handbook/enums.html
https://inspiredwebdev.com/find-replace-in-array/
https://redux.js.org/tutorials/essentials/part-6-performance-normalization

### Related PRs

## Next steps
- Admin permissions: limiting the ability to review requests to admins and giving the ability to lookup other user info by id
- Front-end updates to reflect designs
- Combine purchase requests and inventory updates into an inventory history to show on the inventory page
- Add functionality to upload and display receipt images
- Error messaging/input validation

### Screenshots

https://www.loom.com/share/85caa7e71615475c8591e004c2de7674

CC: @julianrkung


src/lib/redux/inventoryDataSlice.ts
```diff
@@ -66,8 +89,21 @@ const inventoryDataSlice = createSlice({
     addInventory(state, action) {
       state.sitesInventory[action.payload.siteId].siteInventory.push(action.payload);
     },
+    addPurchaseRequest(state, action) {
+      const siteId = action.payload.siteId;
+      delete action.payload.siteId;
+      state.sitesInventory[siteId].purchaseRequests.push(action.payload);
+    },
+    // NOTE: name conflicts with request.js
```
wangannie: Not really sure how to handle this -- should I rename to avoid confusion?

src/lib/airtable/request.js
```diff
@@ -646,6 +665,31 @@ export const updateManyInventorys = async (recordUpdates) => {
...
+// NONGENERATED: Review purchase request
+export const reviewPurchaseRequest = async (purchaseRequest) => {
+  try {
+    // Extract details relevant to reviewing the purchase request
+    const reviewData = {
+      reviewerId: purchaseRequest.reviewerId,
+      approvedAt: purchaseRequest.approvedAt,
+      status: purchaseRequest.status,
+    }
+    updatePurchaseRequest(purchaseRequest.id, reviewData);
```
wangannie: I originally had this as a custom endpoint (see backend PR) but switched it to this -- do you think this is fine for this use case?

src/lib/airtable/request.js
```diff
@@ -225,8 +226,26 @@ export const createManyInventorys = async (records) => {
   return Promise.all(createPromises);
 };
 
-export const createPurchaseRequest = async (record) => {
-  return createRecord(Tables.PurchaseRequests, record);
+// NONGENERATED: Create an inventory record (add product to site)
+export const createPurchaseRequest = async (purchaseRequest) => {
```
wangannie: I'm not really confident this warrants a custom endpoint. Thoughts?

src/lib/redux/inventoryData.ts
```diff
@@ -21,5 +22,53 @@ const addInventoryToRedux = (inventory: InventoryRecord): void => {
...
+  const siteInventoryData = state.inventoryData.sitesInventory[getCurrentSite().id];
+  const inventoryRecord = siteInventoryData.siteInventory.find((inv: InventoryRecord) => inv.id == inventoryId) || "";
+  const products = state.inventoryData.products;
+  if (inventoryRecord) {
+    return products[inventoryRecord.productId];
+  }
+  return EMPTY_PRODUCT;
+}
+
+const updateCurrentQuantityInRedux = (inventoryId: string, newQuantity: number) : void => {
```
julianrkung: maybe `updateInventoryQuantityInRedux`? Keeps it consistent with `updatePurchaseRequestInRedux` and the function's name now shows what it's updating

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -66,8 +89,25 @@ const inventoryDataSlice = createSlice({
...
+    updatePurchaseRequest(state, action) {
+      const siteId = action.payload.siteId;
+      delete action.payload.siteId;
+      const index = state.sitesInventory[siteId].purchaseRequests.findIndex(pr => pr.id === action.payload.id);
+      state.sitesInventory[siteId].purchaseRequests[index] = action.payload;
+    },
+    updateInventoryQuantity(state, action) {
+      const siteId = action.payload.siteId;
+      const inventoryIndex = state.sitesInventory[action.payload.siteId].siteInventory.findIndex(inv => inv.id === action.payload.inventoryId);
```
julianrkung: NIT: use siteId instead of `action.payload.siteId` since you already have a var for it

src/screens/Inventory/CreatePurchaseRequest.tsx
```diff
@@ -0,0 +1,109 @@
...
+      backgroundColor: '#F7F9FC',
+      height: 130,
+      border: `3.5px dashed ${theme.palette.divider}`,
+      radius: '6px',
+    },
+});
+
+interface CreatePurchaseRequestProps extends RouteComponentProps {
+  classes: { content: string; cameraButton: string; };
+  location: any;
```
julianrkung: Side note (probably not for this PR): We should do a refactor soon to wipe away `any` usage for location. It's preventing us from typing out what's getting passed as props via react-router and therefore makes the code significantly harder to read and maintain IMO.

src/screens/Inventory/InventoryProfile.tsx
```diff
@@ -23,20 +23,28 @@ const styles = (theme: Theme) =>
...
+      <Button label={"Purchase"}/>
+    </Link>
+  );
+};
 
 function InventoryProfile(props: InventoryProps) {
   const { classes, products } = props;
-  const inventoryItem: InventoryRecord = props.location.state.inventoryItem;
+  const inventory: InventoryRecord = props.location.state.inventory;
+  const product = products[inventory.productId];
```
julianrkung: use the `getProductByInventoryId` you made here

src/lib/airtable/request.js
julianrkung: We might also want to set the precedent that "Purchase requests must have been persisted to Airtable before they can be approved" (The only workflow that this would cut off is an admin submitted a purchase request and self-approving it).

We might want to do this because the workflow might get too complicated for the workflow of:
1. Offline inventory created
2. While still offline, a purchase request is created
3. While still offline, that same purchase request (which doesn't have a valid id) gets reviewed

src/lib/airtable/request.js
julianrkung: I don't think this warrants a custom endpoint. In the event that we want to do admin authentication (checking the `reviewerId` is an admin, which we should do) on the backend, We can write it in the `write` access resolver for purchase requests.

Details on how to do that with airlock / access resolvers is on: [this](https://share.descript.com/view/mu2V9m33Qa4) demo at around 8:46. If you wanna skip ahead via the transcript, you can CMD+F "So if your rules differ depending on whether or not your users are reading or writing to a base"

src/screens/Inventory/PurchaseRequest.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <BaseScrollView>
+        <div className={classes.content}>
+          <Typography variant="h3">{`Purchase Request for ${product.name}`}</Typography>
+          <Typography variant="body1">{`Request status: ${purchaseRequest.status}`}</Typography>
+          <Typography variant="body1">{`Amount purchased ${purchaseRequest.amountPurchased} ${product.unit}(s)`}</Typography>
+          <Typography variant="body1">{`Amount spent ${purchaseRequest.amountSpent} ks`}</Typography>
+          <Typography variant="body1">{`Created at ${purchaseRequest.createdAt}`}</Typography>
+          <Typography variant="body1">{`Submitted by ${purchaseRequest.requesterId}`}</Typography>
+          {purchaseRequest.notes && <Typography variant="body1">{`Notes: ${purchaseRequest.notes}`}</Typography>}
+          {/* TODO: lookup user by id */}
```
julianrkung: Jsut so I know: Is the lookup to check if they're admins?

src/screens/Inventory/PurchaseRequests.tsx
```diff
@@ -0,0 +1,58 @@
...
+          <Typography variant="h3">Approved</Typography>
+          {siteInventoryData.purchaseRequests.filter(pr => pr.status == PurchaseRequestStatus.Approved).map((purchaseRequest: PurchaseRequestRecord) =>  (
+            <Link key={purchaseRequest.id} to={{ pathname: `/inventory/purchase-requests/purchase-request`, state: { purchaseRequest: purchaseRequest }}}>
+              <PurchaseRequestCard key={purchaseRequest.id} purchaseRequest={purchaseRequest}/>
+            </Link>
+          ))}
+          <Typography variant="h3">Denied</Typography>
+          {siteInventoryData.purchaseRequests.filter(pr => pr.status !== PurchaseRequestStatus.Denied).map((purchaseRequest: PurchaseRequestRecord) =>  (
+            <Link key={purchaseRequest.id} to={{ pathname: `/inventory/purchase-requests/purchase-request`, state: { purchaseRequest: purchaseRequest }}}>
+              <PurchaseRequestCard key={purchaseRequest.id} purchaseRequest={purchaseRequest}/>
```
julianrkung: abstracting this out into a helper function inside of this component would be good (since you're doing the same thing 3 times, 1 for each pending ,approved, denied).

src/screens/Inventory/components/InventoryCard.tsx
```diff
@@ -24,25 +24,21 @@ const styles = (theme: Theme) =>
...
 
 function InventoryCard(props: InventoryCardProps) {
-  const { classes, inventoryItem, products} = props;
+  const { classes, inventory, products} = props;
   return (
     <Card className={classes.cardContainer}>
       <div className={classes.cardContent}>
-        <Typography variant="body1">{products[inventoryItem.productId]?.name}</Typography>
-        <Typography variant="body1" color="textSecondary">Last Updated: 00.00</Typography>
+        <Typography variant="body1">{products[inventory.productId]?.name}</Typography>
```
julianrkung: use the `getProductByInventoryId` you made here

src/lib/airtable/request.js
```diff
@@ -646,6 +665,31 @@ export const updateManyInventorys = async (recordUpdates) => {
   return Promise.all(updatePromises);
 };
 
+// NONGENERATED: Review purchase request
+export const reviewPurchaseRequest = async (purchaseRequest) => {
```
julianrkung: TBH don't know if i really like the name `reviewPurchaseRequest`. Maybe `updatePurchaseRequestAndInventory`? I think my main concern is just that "review" doesn't seem like an action that should persist something.

src/lib/utils/inventoryUtils.ts
```diff
@@ -1,4 +1,11 @@
+import { InventoryRecord } from "../airtable/interface";
+
 // TODO @wangannie replace with robust id generation
 export const generateOfflineInventoryId = (): string => {
     return (Math.floor(Math.random() * 1000)).toString();
 }
+
+// Calculate when an inventory record was last updated
+export const lastUpdated = (inventory: InventoryRecord) => {
```
julianrkung: NIT: generally we want verb names. `getLastUpdated` or `calculateLastUpdated` or something along those lines?

julianrkung (commented):
Left some comments. I haven't tried out workflow yet. Will do that on next iteration!

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -48,6 +53,24 @@ export const EMPTY_INVENTORY_UPDATE: InventoryUpdateRecord = {
   period: 0,
 };
 
+export enum PurchaseRequestStatus {
+  Approved = "Approved",
+  Denied = "Denied",
```
julianrkung: Lets follow the all-caps enum convention we have in codebase already https://github.com/calblueprint/meepanyar/blob/main/src/lib/i18next/translator.ts#L8

src/lib/redux/inventoryDataSlice.ts
julianrkung: We're probably fine here especially if it's behind a `setter` function. If anything, we could probably use namespaces to deal with it?

src/screens/Inventory/PurchaseRequest.tsx
wangannie: Oh oops lookup here is referring to eventually looking up the user's username, since currently I just have it displaying their ID. I guess yeah this would depend on admin access to be able to see a full list of users?

src/lib/redux/inventoryData.ts
```diff
@@ -1,6 +1,19 @@
-import { InventoryRecord, SiteRecord } from '../airtable/interface';
-import { addInventory, saveInventoryData } from './inventoryDataSlice';
-import { store } from './store';
+import { createSelector } from '@reduxjs/toolkit';
+import { InventoryRecord, ProductRecord, PurchaseRequestRecord, SiteRecord } from '../airtable/interface';
+import { addInventory, addPurchaseRequest, EMPTY_INVENTORY, EMPTY_PRODUCT, saveInventoryData, setCurrInventoryId, updateInventoryQuantity, updatePurchaseRequest } from './inventoryDataSlice';
+import { getCurrentSite } from './siteData';
+import { RootState, store } from './store';
+
+export const currentSiteSelector = (state: RootState) => state.siteData.currentSite;
```
julianrkung: Put the `currentSiteSelector` in the `siteData.ts`

While at it, could you put a TODO to remove the `getCurrentSite` getter (we can move all of `siteData` to selectors / entities in a separate PR)?

src/lib/redux/inventoryData.ts
```diff
@@ -1,6 +1,19 @@
...
+export const sitesInventorySelector = (state: RootState) => state.inventoryData.sitesInventory;
+export const allProductsSelector = (state: RootState) => state.inventoryData.products;
+export const currentInventoryIdSelector = (state: RootState) => state.inventoryData.currentInventoryId;
+
+export const currentSiteInventoryDataSelector = createSelector(currentSiteSelector, sitesInventorySelector, (currentSite, sitesInventory) => sitesInventory[currentSite.id]);
+export const currentSiteInventorySelector = createSelector(currentSiteInventoryDataSelector, (siteInventoryData) => siteInventoryData.siteInventory);
+export const currentSitePurchaseRequestsSelector = createSelector(currentSiteInventoryDataSelector, (siteInventoryData) => siteInventoryData.purchaseRequests);
+export const currentInventorySelector = createSelector(currentSiteInventorySelector, currentInventoryIdSelector , (currentSiteInventory, currentInventoryId) => currentSiteInventory.find(inv => inv.id == currentInventoryId) || EMPTY_INVENTORY);
```
julianrkung: Triple equal instead of double equal for `inv.id === currentInventoryId`

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -22,11 +25,13 @@ export type ProductIdString = string;
 interface inventoryDataSliceState {
   products: Record<ProductIdString, ProductRecord>;
   sitesInventory: Record<SiteIdString, SiteInventoryData>;
+  currentInventoryId: string,
```
julianrkung: maybe `currentInventoryItemId`? Not sure if that's better, but it did take me a little to make sure this pointed to an inventory item. Drawback to this is that the Airtable table is literally called `Inventory`, so i'm fine leaving the way it is. Your call on this one

Current naming convention makes `currentInventorySelector` ambiguous (unclear by name if it should return the entire inventory or the current inventory item).

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -1,18 +1,21 @@
...
+import { createEntityAdapter, createSlice, EntityState } from '@reduxjs/toolkit';
 import { InventoryRecord, InventoryUpdateRecord, ProductRecord, PurchaseRequestRecord } from '../airtable/interface';
 
+const inventoryUpdatesAdapter = createEntityAdapter<InventoryUpdateRecord>();
+
 export interface SiteInventoryData {
   siteInventory: InventoryRecord[],
   purchaseRequests: PurchaseRequestRecord[],
-  inventoryUpdates: InventoryUpdateRecord[],
+  inventoryUpdates: EntityState<InventoryUpdateRecord>,
```
julianrkung: Is there a reason that only `inventoryUpdates` is an `EntityAdapter`? Will there be a refactor in the future to make the rest `EntityState`s?

(I'm good with just this structure as long as there'll be a follow-up PR to make the rest `EntityAdapter`s

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -60,14 +89,35 @@ const inventoryDataSlice = createSlice({
...
+    addPurchaseRequest(state, action) {
+      const siteId = action.payload.siteId;
+      delete action.payload.siteId;
+      state.sitesInventory[siteId].purchaseRequests.push(action.payload);
+    },
+    updatePurchaseRequest(state, action) {
+      const siteId = action.payload.siteId;
+      delete action.payload.siteId;
+      const index = state.sitesInventory[siteId].purchaseRequests.findIndex(pr => pr.id === action.payload.id);
+      state.sitesInventory[siteId].purchaseRequests[index] = action.payload;
```
julianrkung: NIT: just for defensive code, I'd like to check the index here is not -1 (I'm not sure if that's possible. Ex: Offline requests persists right before this method is called, so `action.payload.id` is an offlineId but `pr.id` has Airtable IDs).

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -60,14 +89,35 @@ const inventoryDataSlice = createSlice({
...
+    },
+    updateInventoryQuantity(state, action) {
+      const siteId = action.payload.siteId;
+      const inventoryIndex = state.sitesInventory[siteId].siteInventory.findIndex(inv => inv.id === action.payload.inventoryId);
```
julianrkung: Same comment as ^. Defensive code, check the index as a safety measure

src/screens/Inventory/AddInventory.tsx
```diff
@@ -93,11 +90,4 @@ function AddInventory(props: AddInventoryProps) {
   );
 }
 
-const mapStateToProps = (state: RootState) => ({
```
julianrkung, line 96: I see we're removing `mapStateToProps` in favor of `useSelector`. I'm fine with this, but is there a general rule where we'd want to use `mapStateToProps` over `useSelector`? Or do we want to just migrate to `useSelector` longterm (assuming we're using functional components)

src/screens/Inventory/CreatePurchaseRequest.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const [amountPurchased, setAmountPurchased] = useState(0.0);
+  const [amountSpent, setAmountSpent] = useState(0.0);
+  const [notes, setNotes] = useState("");
+
+  const handleAmountPurchasedInput = (event: React.ChangeEvent<{ value: unknown }>) => {
+    setAmountPurchased(parseFloat(event.target.value as string));
+  }
+
+  const handleAmountSpentInput = (event: React.ChangeEvent<{ value: unknown }>) => {
+    setAmountSpent(parseFloat(event.target.value as string));
```
julianrkung: We should make sure to restrict these to numbers. I put in letters and it let me submit and the amonut became a `NaN`

image

src/lib/airtable/request.js
```diff
@@ -646,6 +667,23 @@ export const updateManyInventorys = async (recordUpdates) => {
   return Promise.all(updatePromises);
 };
 
+// NONGENERATED: Approve/deny purchase request and update inventory qty if approved
+export const updatePurchaseRequestAndInventory = async (purchaseRequest) => {
```
julianrkung: This is now just `updatePurchaseRequest` right? Since inventory update is in `createPurchaseRequest`

src/lib/airtable/request.js
```diff
@@ -225,8 +225,29 @@ export const createManyInventorys = async (records) => {
   return Promise.all(createPromises);
 };
 
-export const createPurchaseRequest = async (record) => {
-  return createRecord(Tables.PurchaseRequests, record);
+// NONGENERATED: Create a Purchase Request and update the inventory's current qty (regardless of approval status)
+export const createPurchaseRequest = async (purchaseRequest) => {
```
julianrkung: now should be called `createPurchaseRequestAndUpdateInventory`?

src/screens/Inventory/PurchaseRequests.tsx
```diff
@@ -0,0 +1,59 @@
...
+// TODO @wangannie: address empty state
+function PurchaseRequests (props: PurchaseRequestsProps) {
+    const { classes } = props;
+    const purchaseRequests = useSelector(currentSitePurchaseRequestsSelector);
+
+    const getPurchaseRequestsSection = (status: PurchaseRequestStatus) => (
+      <div>
+        <Typography variant="body1">{status}</Typography>
+        {purchaseRequests.filter(pr => pr.status == status).map((purchaseRequest: PurchaseRequestRecord) =>  (
+          <Link key={purchaseRequest.id} to={{ pathname: `/inventory/purchase-requests/purchase-request`, state: { purchaseRequest }}}>
```
julianrkung, line 30: Are we fine passing purchaseRequest here via react-router? I'm good with it, just want to make sure it's intentional. A `currentPurchaseRequestId` might seem annoying to deal with

julianrkung (commented):
Left some comments! Code looks a lot cleaner with the selectors, good job :)

src/screens/Inventory/CreatePurchaseRequest.tsx
wangannie: I made it default to 0 if NaN temporarily, but tabling this for a later PR to address app-wide error states and input validation.

src/screens/Inventory/PurchaseRequests.tsx
wangannie: Yeah I figured this flow would be fine (at least with the current use cases) since we don't really deal with editing a purchase request (besides reviewing, in which case the reviewer is sent back to the main list anyways). I'll probably re-evaluate once we have more finalized designs and user flows. If you think it's likely we're going to standardize  _everything_ to keeping an currentXXXId in state, I'm happy to refactor this too.

src/lib/redux/inventoryDataSlice.ts
wangannie: Yeah I definitely understand the confusion here, but I think I'm trying to stick to this standard:
- `siteInventory`: list of inventory items that a site has
- `inventory`: a single inventory 'item', aligning with the Airtable table

So this way, for example, `selectCurrentInventory` represents the current inventory 'item', and `selectCurrentSiteIventory` = list of inventory items at a site.

Let me know what you think. I'm not super attached to my naming scheme so if you think it's better to shift to 'item' I'd be open to considering it.

src/screens/Inventory/AddInventory.tsx
wangannie: Yeah I think in general I ended up moving all of them over to use selectors because I was always grabbing the same kinds of info, which made it a lot cleaner and more readable to switch to selectors. Also, it got pretty nasty with all the nested params I needed for typescript to agree with `mapStateToProps` (eg `const mapStateToProps = (state: RootState, ownProps: { location: { state: { purchaseRequest: { inventoryId: string; }; }; }; }) => ({` from [here](https://github.com/calblueprint/meepanyar/pull/60/files/afe73e075a92a0ef25939bb4b0eb687b55748af2#diff-bd8ec877cd98f9458d9ae9715e5c346b6a69661fc3cd5671b522b40f1f4ad27fR61-R63).

In general, I think the Redux docs do a better job explaining this, but they did mention

> The existing connect API still works and will continue to be supported, but the hooks API is simpler and works better with TypeScript. ([Source](https://react-redux.js.org/api/hooks#hooks))

src/lib/redux/inventoryDataSlice.ts
wangannie: wOoooO this was fixed using `createEntityAdapter`'s built-in magic ✨

src/lib/airtable/request.js
```diff
@@ -226,30 +226,27 @@ export const createManyInventorys = async (records) => {
...
-    const resp = await fetch(`${process.env.REACT_APP_AIRTABLE_ENDPOINT_URL}/purchase-request/create`, {
-      method: 'POST',
-      headers: {
-        'Content-Type': 'application/json'
-      },
-      body: JSON.stringify(purchaseRequest)
-    })
-    console.log('Response for purchase request: ', resp);
-    await resp.json().then(data => purchaseRequestId = data.id);
+    delete purchaseRequest.id; // Remove the id field to add to Airtable
```
julianrkung, line 246: Can you make a note that this doesn't handle offline yet? In particular I don't believe it handles the flow:
1. An inventory item was created while offline
2. While still offline, a purchase request was created for that inventory item (in this case the inventory item doesn't have an ID)

Let me know if i'm off the mark on that one. We can do that in a future PR

src/lib/redux/inventoryDataSlice.ts
```diff
@@ -2,19 +2,55 @@
 /* eslint-disable no-param-reassign */
 import { createEntityAdapter, createSlice, EntityState } from '@reduxjs/toolkit';
 import { InventoryRecord, InventoryUpdateRecord, ProductRecord, PurchaseRequestRecord } from '../airtable/interface';
+import { setCurrSite } from './siteDataSlice';
+import { RootState } from './store';
 
 const inventoryUpdatesAdapter = createEntityAdapter<InventoryUpdateRecord>();
+const productsAdapter = createEntityAdapter<ProductRecord>();
+const siteInventoryAdapter = createEntityAdapter<InventoryRecord>();
+const purchaseRequestsAdapter = createEntityAdapter<PurchaseRequestRecord>();
```
julianrkung: This entire setup and file looks really nice, really good work :)

I'll work on refactoring `customerData` and `siteData` to look similar to this

src/screens/Inventory/PurchaseRequest.tsx
```diff
@@ -28,23 +28,25 @@ const styles = (theme: Theme) =>
 interface PurchaseRequestsProps extends RouteComponentProps {
   classes: { content: string; section: string; };
   location: any;
-  product: ProductRecord;
+  // product: ProductRecord;
```
julianrkung: rm commented line if not needed

src/screens/Inventory/PurchaseRequest.tsx
```diff
@@ -28,23 +28,25 @@ const styles = (theme: Theme) =>
...
 }
 
 
 function PurchaseRequest (props: PurchaseRequestsProps) {
-  const { classes, product } = props;
+  const { classes } = props;
   const purchaseRequest: PurchaseRequestRecord = props.location.state.purchaseRequest;
+  const product = useSelector((state: RootState) => selectProductById(state, selectCurrentSiteInventoryById(state, purchaseRequest.inventoryId)?.productId || ""))|| EMPTY_PRODUCT;
```
julianrkung: Is there any way this can be split into 2 lines? The line is 180 characters and is kinda hard to read

julianrkung: Would you mind doing some redirecting if someone navigates to this URL by typing it in (maybe just redirecting them to inventoryMain)? I'll be doing the equivalent handling in the customer-related pages.

src/screens/Inventory/InventoryProfile.tsx
```diff
@@ -31,8 +32,8 @@ const getPurchaseRequestButton = () => (
 
 function InventoryProfile(props: InventoryProps) {
   const { classes } = props;
-  const inventory = useSelector(currentInventorySelector);
-  const product = useSelector(currentInventoryProductSelector);
```
julianrkung: Would you mind doing some redirecting if someone navigates to this URL by typing it in (maybe just redirecting them to inventoryMain) when these aren't set yet? I'll be doing the equivalent handling in the customer-related pages.

src/screens/Inventory/CreatePurchaseRequest.tsx
```diff
@@ -39,19 +39,20 @@ function CreatePurchaseRequest(props: CreatePurchaseRequestProps) {
   const { classes } = props;
   const history = useHistory();
   const userId = useSelector(userIdSelector);
-  const inventory = useSelector(currentInventorySelector);
-  const product = useSelector(currentInventoryProductSelector);
```
julianrkung: Would you mind doing some redirecting if someone navigates to this URL by typing it in (maybe just redirecting them to inventoryMain) when these aren't set yet? I'll be doing the equivalent handling in the customer-related pages.

src/screens/Inventory/components/PurchaseRequestCard.tsx
```diff
@@ -41,7 +38,8 @@ const styles = (theme: Theme) =>
 
 // TODO: sort by creation date/status
 function PurchaseRequestCard(props: PurchaseRequestCardProps) {
-  const { classes, product, amountPurchased, createdAt, requesterId, status } = props;
+  const { classes, inventoryId, amountPurchased, createdAt, requesterId, status } = props;
+  const product = useSelector((state: RootState) => selectProductById(state, selectCurrentSiteInventoryById(state, inventoryId)?.productId || "")) || EMPTY_PRODUCT;
```
julianrkung: Can we separate this out to 2 lines? This line is 165 characters. Maybe have getting the currentSiteInventoryId the line before?

julianrkung (commented):
Really small NITs at this point.

Couldn't check out the actual flow (only looked at the code) because when I tried to log in using backend `main` I got the following error:
image

julianrkung (commented):
The error resolved itself after I cleared site data (devtools --> application --> storage --> clear site data). I checked the workflow and everything looked good on that end. I'll approve after the changes from previous PR are done

src/screens/Inventory/CreatePurchaseRequest.tsx
wangannie: `location` is conveniently phased out with selectors :~)

src/screens/Inventory/PurchaseRequest.tsx
wangannie: Added redirects and updated the PR description with a testable workflow!

src/lib/airtable/patches/patch-schema.patch
```diff
@@ -1,5 +1,5 @@
 diff --git a/src/lib/airtable/schema.js b/src/lib/airtable/schema.js
-index 7c751ee..e71d320 100644
```
wangannie, line 2: Not sure if this is necessary/what it's actually doing, but this was generated when I ran the git diff.

julianrkung (approved):
Followed the review instructions and everything looks good. I also switched sites and went to `/inventory/purchase-requests/purchase-request` to make sure that the `currentInventoryId` was reset and I was redirected to inventory main. Everything looks good :). You're the best! This was built out so quickly and with such high quality I'm excited for future PRs HAHA

We can do this in a future PR, but this Purchase Request column should probably just be called "Reviewed At" since it gets populated even when I deny
image

PR #62, opened 2021-03-26
Customer props passed via react-router and edge cases

julianrkung:

# What's new in this PR
In this PR: 
1. we remove `currentCustomer` from redux and pass `customer` to `CustomerProfile` and `EditCustomer` via react-router. This was because we decided to not keep state `currentCustomer` state inside of redux and instead pass it.
2. Fixed a race condition bug where we would navigate to `customerMain` before `editCustomer` finished persisting to redux and so `customerMain` would show the outdated customer instead of the edited customer
3. Took care of edge conditions where a customer prop isn't passed when going to `customerProfile` and `EditProfile`.

### Main functionality
1. Log into the app
2. Edit a customer
3. Go back to the `CustomerMain` page and ensure that customer is edited
4. Make sure the customer is edited appropriately on airtable and an associated `Customer Update` was created for the change.

### Edge cases
1. Log into the app and navigate to `/customers/customer` via the URL bar and make sure you get redirect to the `/customers` endpoint without error
2. navigate to `/customers/customer/edit` via the URL bar and make sure you get redirect to the `/customers` endpoint without error

## Relevant Links
I took basically all this PR from Annie's PR suggestions https://github.com/calblueprint/meepanyar/compare/tiffany%2Fedit-customer...annie%2Fedit-customer-suggestions?diff=split&expand=1

### Online sources

### Related PRs

## Next steps

### Screenshots

CC: @julianrkung

[//]: # 'This tags Julian as a default. Feel free to change, or add on anyone who you should be in on the conversation.'****

### DIFF
--- a/src/screens/Customers/EditCustomer.tsx
+++ b/src/screens/Customers/EditCustomer.tsx
@@ -2,7 +2,7 @@ import { createStyles, FormControl, InputLabel, Theme, MenuItem, Select } from '
 import { withStyles } from '@material-ui/core/styles';
 import React, { useState } from 'react';
 import { connect } from 'react-redux';
-import { RouteComponentProps, useHistory } from 'react-router-dom';
+import { RouteComponentProps, useHistory, Redirect } from 'react-router-dom';
 import BaseScreen from '../../components/BaseComponents/BaseScreen';
 import Button from '../../components/Button';
 import Checkbox from '../../components/Checkbox';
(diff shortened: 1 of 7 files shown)


src/screens/Customers/EditCustomer.tsx
```diff
@@ -49,6 +52,11 @@ function EditCustomer(props: EditCustomerProps) {
   const [meterNumber, setMeterNumber] = useState(currentCustomer.meterNumber);
   const [customerInactive, setCustomerInactive] = useState(!currentCustomer.isactive);
 
+  // Redirect to main customers page if no customer was passed
+  if (currentCustomer === EMPTY_CUSTOMER) {
```
julianrkung, line 56: `EMPTY_CUSTOMER` was used here because the `useState` use `currentCustomer`. I also couldn't do this if clause before all the `useState` calls because any return statements need to happen after all the hooks execute (react yelled at me)

julianrkung:
Closing because decided to go with `setCurrentCustomerId` in redux after talking with @wangannie offline. We decided this because it's cleaner than passing the customer as a prop in react router in most cases.

Race conditions and edge conditions will be dealt with in the [add customerData PR](https://github.com/calblueprint/meepanyar/pull/56)

PR #64, opened 2021-03-31
Customer Data uses Entity Adapter and Selectors

julianrkung:

# What's new in this PR
- This PR makes `customerDataSlice` consistent with `inventoryDataSlice` from this [PR](https://github.com/calblueprint/meepanyar/pull/60).
- Makes the customerData use an entity adapter
- Create selectors to be able to select from `customerData`
- Refactored so that customer-related screens use selectors

### Adding a Customer
- Create a customer for a site
- Ensure in the redux store that the customer has a proper Airtable record ID
- Check the backend that the customer was properly created in the correct site

### Editing a Customer
- Edit the customer (can be the customer created from the "Adding a Customer" section)
- Ensure the edit is persisted on the frontend (Customer Main, Customer Profile) and redux store
- Ensure that a "Customer update" was created in Airtable and the customer record is properly edited

### Redirects
- Change the site to reset the "CurrentCustomerId" value in redux
- Go to `/customers/customer` via the URL bar and make sure it redirects you to `/customers`
- Go to `/customers/customer/edit` via the URL bar and make sure it redirects you to `/customers`

### Online sources
Resources from Annie's PR: https://github.com/calblueprint/meepanyar/pull/60

### Related PRs
https://github.com/calblueprint/meepanyar/pull/60

## Next steps
- Found a bug with the customer search. It doesn't handle deleting then re-entering different characters well.
  - Ex: In "East Asian Library" site, you can search "Al..." to narrow the search, then delete the "l" and add a space to attempt to search for the "A new customer" customer, but nothing will show up.
- Refactor SiteData to also use selectors 

### Screenshots

CC: @julianrkung

### DIFF
--- a/src/lib/redux/customerData.ts
+++ b/src/lib/redux/customerData.ts
@@ -1,32 +1,29 @@
 import { CustomerRecord } from '../airtable/interface';
-import { addCustomer, setCurrentCustomerId, editCustomer, EMPTY_CUSTOMER } from './customerDataSlice';
+import { addCustomer, setCurrentCustomerId, editCustomer, selectCustomerById } from './customerDataSlice';
 import { getCurrentSiteId } from './siteData';
-import { store } from './store';
+import { RootState, store } from './store';
+import { createSelector } from '@reduxjs/toolkit';
 
-const addCustomerToRedux = (customer: CustomerRecord): void => {
+export const addCustomerToRedux = (customer: CustomerRecord): void => {
     const siteId = getCurrentSiteId();
     store.dispatch(addCustomer({ siteId, customer }));
 };
 
-const editCustomerInRedux = (customer: any): void => {
-    store.dispatch(editCustomer(customer));
+export const editCustomerInRedux = (customer: any): void => {
+    const customerUpdates = {
+        ...customer,
+        siteId: getCurrentSiteId()
+    }
+    store.dispatch(editCustomer(customerUpdates));
 };
 
-const setCurrentCustomerIdInRedux = (customerId: string): void => {
+export const setCurrentCustomerIdInRedux = (customerId: string): void => {
     store.dispatch(setCurrentCustomerId(customerId))
 }
 
-const getCurrentCustomer = (): CustomerRecord | undefined => {
-    // TODO: When changing to entity adapter, we can speed this up / use a Selector instead
-    const customers = getAllCustomersInSite();
-    const currentCustomerId = store.getState().customerData.currentCustomerId;
-    return customers.find(customer => customer.id === currentCustomerId);
-}
+export const selectCurrentCustomerId = (state: RootState) => state.customerData.currentCustomerId;
 
-const getAllCustomersInSite = (): CustomerRecord[] => {
-    const siteId = getCurrentSiteId();
-    const siteIdsToCustomers = store.getState().customerData.siteIdsToCustomers[siteId] || {};
-    return Object.values(siteIdsToCustomers);
-}
-
-export { setCurrentCustomerIdInRedux, addCustomerToRedux, getAllCustomersInSite, editCustomerInRedux, getCurrentCustomer };
+export const selectCurrentCustomer = createSelector(
+    selectCurrentCustomerId, 
+    store.getState, 
+    (currentCustomerId, state) => selectCustomerById(state, currentCustomerId));
--- a/src/lib/redux/customerData.ts
+++ b/src/lib/redux/customerData.ts
@@ -9,7 +9,7 @@ export const addCustomerToRedux = (customer: CustomerRecord): void => {
     store.dispatch(addCustomer({ siteId, customer }));
 };
 
-export const editCustomerInRedux = (customer: any): void => {
+export const editCustomerInRedux = (customer: Partial<CustomerRecord>): void => {
     const customerUpdates = {
         ...customer,
         siteId: getCurrentSiteId()
--- a/src/lib/redux/customerDataSlice.ts
+++ b/src/lib/redux/customerDataSlice.ts
@@ -1,8 +1,9 @@
 /* eslint-disable no-unused-vars */
 /* eslint-disable no-param-reassign */
-import { createSlice } from '@reduxjs/toolkit';
+import { createEntityAdapter, createSlice, EntityState } from '@reduxjs/toolkit';
 import { CustomerRecord, SiteId } from '../airtable/interface';
 import { setCurrSite } from './siteDataSlice';
+import { RootState } from './store';
 
 export const EMPTY_CUSTOMER: CustomerRecord = {
     id: '',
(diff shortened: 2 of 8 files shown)


src/lib/redux/customerData.ts
```diff
@@ -1,32 +1,29 @@
...
-const addCustomerToRedux = (customer: CustomerRecord): void => {
+export const addCustomerToRedux = (customer: CustomerRecord): void => {
     const siteId = getCurrentSiteId();
     store.dispatch(addCustomer({ siteId, customer }));
 };
 
-const editCustomerInRedux = (customer: any): void => {
-    store.dispatch(editCustomer(customer));
+export const editCustomerInRedux = (customer: any): void => {
+    const customerUpdates = {
```
wangannie, line 13: For `updatePurchaseRequestInRedux` I ended up using `Partial<PurchaseRequestRecord>` as the parameter type (see code [here](https://github.com/calblueprint/meepanyar/pull/60/files#diff-073d3298a1ef44614cb4a2cf9fc92f6cd763e00c3ada134e68b7b8d0b9cb5bf8R58-R65)). It's not essential but maybe something to consider for consistency

src/lib/redux/customerDataSlice.ts
```diff
@@ -22,15 +23,26 @@ export const EMPTY_CUSTOMER: CustomerRecord = {
...
+    selectById: selectCustomerById,
+    selectIds: selectCustomerIds
+} = customersAdapter.getSelectors(
+    (state: RootState) => state.customerData.customers[state.siteData.currentSite.id]);
+
 // TODO: @julianrkung, We should eventually make this a Record<SiteId, Record<CustomerId, CustomerRecord>> map for efficiency
 // We can't do this yet because customers created while offline do not receive an id yet
 interface customerDataSliceState {
-    siteIdsToCustomers: Record<SiteId, CustomerRecord[]>;
+    customers: Record<SiteId, EntityState<CustomerRecord>>;
```
wangannie: This naming is a little misleading. To match the naming in inventory (`sitesInventory`) maybe `sitesCustomers` ?? Though it doesn't sound the best.

src/screens/Customers/CustomerMain.tsx
```diff
@@ -73,29 +73,25 @@ const labels: FilterByLabel = {
...
   const [filteredCustomers, setFilteredCustomers] = useState<CustomerRecord[]>([]);
   const [filteredCustomersAlt, setFilteredCustomersAlt] = useState<CustomerRecord[]>([]);
-  const [fullCustomers, setFullCustomers] = useState<CustomerRecord[]>([]);
   const [allCustomersTrie, setAllCustomersTrie] = useState<TrieTree<CustomerRecord>>(new TrieTree('name'));
   const [sortBy, setSortBy] = useState<SortBy>(SortBy.NAME);
   const [filterBy, setFilterBy] = useState<FilterBy>(FilterBy.ACTIVE_STATUS);
   const [sortAndFilter, setSortAndFilter] = useState<string[]>([])
   const [filterLabels, setFilterLabels] = useState<string[]>(labels["ACTIVE_STATUS"]);
   const [searchValue, setSearchValue] = useState<string>("");
+  const fullCustomers = useSelector(selectAllCustomersArray) || [];
```
wangannie: This breaks when it's `[]` -- you might have to use `[] as CustomerRecord[]` instead

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -53,7 +53,12 @@ interface CustomerProps extends RouteComponentProps {
 }
 
 function CustomerProfile(props: CustomerProps) {
-  const { classes, match, currentSite, customer } = props;
+  const { classes, match, currentSite } = props;
+  const customer : CustomerRecord = useSelector(selectCurrentCustomer) || EMPTY_CUSTOMER;
+
+  if (customer === EMPTY_CUSTOMER) {
```
wangannie: For inventory, I ended up taking out the `|| EMPTY` and just had the redirect if condition be `if (!customer)` which I think makes more sense since there's really no good reason for there to be an empty customer except to satisfy the type (which you can just remove).

src/screens/Customers/EditCustomer.tsx
```diff
@@ -39,9 +39,11 @@ interface EditCustomerProps extends RouteComponentProps {
 }
 
 function EditCustomer(props: EditCustomerProps) {
-  const { classes, currentCustomer, currentSite, user } = props;
+  const { classes, currentSite, user } = props;
   const history = useHistory();
 
+  const currentCustomer = useSelector(selectCurrentCustomer) || EMPTY_CUSTOMER;
```
wangannie, line 43: See my comment in `CustomerProfile`

src/screens/Customers/CustomerMain.tsx
wangannie: Alternatively I think you can add `fullCustomers: CustomerRecord[]` based on what you did in `Home`

src/screens/Customers/EditCustomer.tsx
```diff
@@ -39,9 +39,11 @@ interface EditCustomerProps extends RouteComponentProps {
 }
 
 function EditCustomer(props: EditCustomerProps) {
```
wangannie, line 39: Remove `currentCustomer` from `EditCustomerProps`

src/screens/Customers/EditCustomer.tsx
```diff
@@ -117,7 +126,6 @@ function EditCustomer(props: EditCustomerProps) {
 }
 
 const mapStateToProps = (state: RootState) => ({
-  currentCustomer: getCurrentCustomer() || EMPTY_CUSTOMER,
   currentSite: state.siteData.currentSite || EMPTY_SITE,
   user: state.userData.user || EMPTY_USER,
```
wangannie: I wrote a `selectCurrentUserId` selector for inventory data that you should use here

wangannie (changes requested):
Looking good! I mostly just noted some cleanup things. Also, I saw that `currentCustomer` is still in `siteDataSlice` which I believe is outdated? That might have been missed in the customerData PR.

src/screens/Customers/EditCustomer.tsx
julianrkung: Can't do it here because `useState` lines below rely on `currentCustomer` being non-null. I don't think React lets you exit before all hooks have run, so I think `EMPTY_CUSTOMER` is necessary in this particular case

src/screens/Customers/CustomerMain.tsx
julianrkung: Did the latter, thanks!

src/screens/Customers/EditCustomer.tsx
julianrkung: Good catch, thanks

src/lib/redux/customerData.ts
julianrkung: Good call, thanks!

src/lib/redux/customerDataSlice.ts
julianrkung: I changed it to `sitesCustomers` for consistency. I'd really prefer the name change of `siteIdsToCustomers` in a future refactor if you're down since I agree name doesn't sound great

wangannie: Ah yes I'm down

wangannie (approved):
Changes look good to me!

PR #66, opened 2021-04-01
[Inventory] Add Inventory Updates functionality 

wangannie:

# What's new in this PR
- Added basic functionality to create an inventory update
- Trigger an inventory update when inventory is added to a site (to track the user id and other details for recordkeeping)
- Added calculation for the `lastUpdated` timestamp based on the most recent purchase request OR inventory update
- Updated the way we calculate and format date/time strings to be sensitive to location and time zone.

## How to review

**Add inventory**
1. On the main inventory screen, select the + button to add a new inventory item
2. Select any item, input a quantity, and submit.
3. Check both Airtable and Redux to see that an Inventory Update was created for the new item.

**Create an inventory update**
1. Select any inventory item from the main list.
2. Click **Update**
3. Enter any quantity (note: we don't have robust edge case handling, so no need to try to break this) and click Update
4. Check that the quantity in the inventory profile is updated, as well as in the inventory record, and that an inventory update was created (check in both Airtable and Redux).
5. Check that a new InventoryUpdateCard shows up on the profile page at the **top** representing your most recent update.
6. Check that the Last Updated date reflects the approximate time you hit submit.

**Last Updated Date**
1. On an inventory item that already has an inventory update (basically all of them), create a purchase request.
2. Make sure the 'Last Updated' date now reflects the time you submitted the purchase request.

**Test Redirect**
1. Reset the Redux store in the inspector. You may also have to do Clear site data in the application tab as well.
2. Log in as testUser
3. Directly enter http://localhost:3000/inventory/updates/create into the URL bar
4. Make sure it redirects to the main inventory screen

### Online sources
https://momentjs.com/docs/#/get-set/max/
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString

### Related PRs
n/a

## Next steps
- Handle error states and edge case inputs
- Update frontend to reflect the designs

### Screenshots
image
image

CC: @julianrkung


src/lib/airtable/request.js
```diff
@@ -271,6 +274,32 @@ export const createManyPurchaseRequests = async (records) => {
   return Promise.all(createPromises);
 };
 
+// NONGENERATED: Create an Inventory Update and update the inventory's current qty
+// TODO: handle offline workflow of creating inventory updates for inventory
+// that was created offline (no Airtable id).
+export const createInventoryUpdateAndUpdateInventory = async (userId, inventory, updatedAmount ) => {
```
wangannie, line 283: Since this is used in multiple places, I ended up just moving all the `inventoryUpdate` creation work into here. Is that bad practice?

Also this function name is a nightmare and I'm definitely open to suggestions 😅

src/lib/utils/inventoryUtils.ts
```diff
@@ -1,6 +1,25 @@
...
+  // Check because moment will incorrectly return the current time if updateTimes is empty
+  if (updateTimes.length > 0) {
+    return moment.max(updateTimes).toLocaleString();
+  }
+  return 'Unknown'; // TODO @wangannie: address design edge case
+};
+
+// TODO @wangannie: fill in with real period calculation
+export const getCurrentPeriod = (): number => {
+  return moment().month();
```
julianrkung: Just a side note (no action). Periods are defined as `YYYY-MM` because month isn't enough to uniquely identify a period

src/screens/Inventory/components/InventoryUpdateCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+  createdAt: string;
+  userId: string;
+}
+
+function InventoryUpdateCard(props: InventoryUpdateCardProps) {
+  const { classes, inventoryId, updatedQuantity, createdAt, userId } = props;
+  const product =
+    useSelector((state: RootState) =>
+      selectProductById(state, selectCurrentSiteInventoryById(state, inventoryId)?.productId || ''),
+    ) || EMPTY_PRODUCT;
```
julianrkung: This same code is in `PurchaseRequest.tsx`. Is there any way to define this selector in `InventoryData` and import it?

Just to clarify for my understanding: The only difference between this and `selectCurrentInventoryProduct` in `inventoryData` is that this function doesn't rely on `currentInventoryId` right?

src/lib/airtable/request.js
julianrkung: Kinda bad practice because we don't use this convention anywhere else, but as long as you put comments in here about why you're doing it this way I'm down to do this since I see the upside also if it's being used in multiple places.

src/screens/Inventory/CreateInventoryUpdate.tsx
```diff
@@ -0,0 +1,77 @@
...
+  if (!userId || !inventory || !product) {
+    return <Redirect to={'/inventory'} />;
+  }
+
+  // TODO @wangannie: add better edge case handling
+  const handleUpdatedAmount = (event: React.ChangeEvent<{ value: unknown }>) => {
+    setUpdatedAmount(parseFloat(event.target.value as string) || 0);
+  };
+
+  const handleSubmit = async (event: React.MouseEvent) => {
```
julianrkung: This also has the bug where if you click multiple times in quickly, unintended behavior occurs. Did you want to fix that behavior all at once sometime later?

julianrkung (commented):
Looks basically done, really clean and fast work! Left some comments, but only the `InventoryUpdateCard` one really requires any action.

When you ping me again for review, include the patch to request so that I can review that as well quickly before approving

src/lib/utils/inventoryUtils.ts
wangannie: Good catch! I forgot to delete this when I decided to remove period from Airtable

src/screens/Inventory/components/InventoryUpdateCard.tsx
wangannie: Right. I figured this shouldn't rely on currentInventory in case we need a history feed pulling update cards for different inventory items

src/screens/Inventory/CreateInventoryUpdate.tsx
wangannie: Fixed with a loading circle + disabling the button! It's a temporary fix and we'll probably need to check with Ashley on the design, but it prevents double clicks
image

src/components/Button.tsx
```diff
@@ -20,13 +20,14 @@ interface MainButtonProps {
...
   onClick?: (event: React.MouseEvent) => void;
+  loading?: boolean;
 }
 
 function MainButton(props: MainButtonProps) {
   const { classes } = props;
   return (
-    <Button className={classes.button} type="submit" variant="contained" color="primary" disableElevation={true} onClick={props.onClick}>
-      {props.label}
+    <Button disabled={props.loading} className={classes.button} type="submit" variant="contained" color="primary" disableElevation={true} onClick={props.onClick}>
```
julianrkung, line 29: Thanks for taking care of this! :^)

src/lib/utils/inventoryUtils.ts
```diff
@@ -1,6 +1,29 @@
...
+
+  const updateTimes: Moment[] = [];
+  purchaseRequests.map((pr) => updateTimes.push(moment(pr.createdAt)));
+  inventoryUpdates.map((update) => updateTimes.push(moment(update.createdAt)));
+
+  // Check because moment will incorrectly return the current time if updateTimes is empty
+  if (updateTimes.length > 0) {
+    return moment.max(updateTimes).toLocaleString();
+  }
+  return 'Unknown'; // TODO @wangannie: address design edge case
```
julianrkung, line 29: Question: When would this happen (is this just protecting against admins / airtable people deleting the initial / all inventory updates)? Because isn't there an inventory update when the inventory is created? It's not bad, but was wondering if this is just defensive code or if we're expecting an 'Unknown' return under some circumstances

src/screens/Inventory/CreateInventoryUpdate.tsx
julianrkung: Kinda small, but is there a reason for this being `async`? I think this still creates a race condition (probably subtle enough that I can't reproduce) since the `setLoading` is not synchronous with the clicking of the button.

I think chaining the `history.goBack` to `createInventoryUpdateAndUpdateInventory` and making the function not async is fine, is it not?

src/screens/Inventory/components/InventoryUpdateCard.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+function InventoryUpdateCard(props: InventoryUpdateCardProps) {
+  const { classes, inventoryId, updatedQuantity, createdAt, userId } = props;
+  const product = useSelector((state: RootState) => selectProductByInventoryId(state, inventoryId)) || EMPTY_PRODUCT;
+
+  return (
+    <Card className={classes.cardContainer}>
+      <div className={classes.cardContent}>
+        <Typography variant="body1">{`${product.name}, ${updatedQuantity} ${product.unit}(s)`}</Typography>
+        <Typography variant="body1">{`Created at: ${createdAt}`}</Typography>
```
julianrkung: I imagine we want to eventually change this to the timezone of the user. The only one I see that isn't in the appropriate timezone is the one in `createdAt`. There are other dates I think that are unchanged too, so we could change this in a future PR, but since it's quick I'd rather not have to do more work down the line

Screenshot: I created this at 8AM PST 4/3/2021
image

julianrkung (approved):
Looks good! I approved because there's nothing wrong with the workflow and comments were minor. Fix the `createdAt` timezone thing (and potentially the potentially unnecessary async? I think another submit inside the codebase was async as well when I was looking over) and feel free to merge! Thanks for getting this done so quickly, I am running out of compliments to give. HAHAHAHA but I really appreciate you being so on top of it!! :)

src/lib/utils/inventoryUtils.ts
wangannie: Hmm yeah I think it's just because by default moment will incorrectly return the current time if given an empty array, which would be very misleading if that were to ever happen for any reason (but I agree, no specific cases come to mind besides the airtable cases you mentioned). I guess this is just to make sure we don't ever let it just return the current time by default.

src/screens/Inventory/components/InventoryUpdateCard.tsx
wangannie: I ended up reformatting the way we do dates and times entirely by updating the helper function in `momentUtils` to return a string format that is sensitive to both language and timezone. I was worried about different locales having different formats/orderings of month/day/year etc so I think this resolves that issue.

**Myanmar**
image

**San Francisco**
image

**Tokyo (just for contrast)**
image

julianrkung (approved):
Thank you for fixing dates :)

PR #67, opened 2021-04-02
[Inventory] Create new product (inventory type) functionality

wangannie:

# What's new in this PR
In the Add Inventory flow, users can now create new products (inventory types).

## How to review
**Create a new product**
1. On the inventory screen, click the + to open the Add Inventory screen
2. In the dropdown, either start typing or scroll to the bottom of the dropdown to find '+ New Inventory Item'
3. Two new text fields should pop up to enter a name and unit for the new product.
4. Fill in a number for starting amount and submit.
5. Verify that:
  - A new product was created in Airtable and under `inventoryData.products` in Redux
  - A new inventory was created in Airtable and under `inventoryData.sitesInventory.[siteId].siteInventory` in Redux with the correct product id
 
**Offline Ids**
If `await createProduct` fails for any reason, it should generate an 'offline id' for the product which will be used for the inventory item. To test this, try going through the same steps and the new records should be added to redux with integer ids.

### Online sources
https://material-ui.com/components/autocomplete/

### Related PRs
n/a

## Next steps
- Error handling/input validation

### Screenshots
https://www.loom.com/share/af984ae848f34c4c8bac89da317756b2

image

CC: @julianrkung


src/screens/Inventory/AddInventory.tsx
```diff
@@ -61,29 +94,66 @@ function AddInventory (props: AddInventoryProps) {
...
-            getOptionLabel={(productId) => `${products[productId]?.name} (${products[productId]?.unit})`}
-            renderInput={(params) => <MaterialTextField {...params} label="Item" variant="outlined"/>}
+            value={selectedProductId}
+            style={{ marginBottom: 20 }}
             onChange={handleSelectItem}
+            filterOptions={(options, params) => {
+              const filtered = filter(options, params);
+              filtered.push(NEW_PRODUCT_LABEL);
+              return filtered;
+            }}
```
julianrkung, line 126: Based on https://material-ui.com/components/autocomplete/#createfilteroptions-config-filteroptions and the API documentation, I don't think I understand what's going on in this line. I thought you're just pass the `filter` you create to `filterOptions`. It might be the reason it's not working for me?

image

julianrkung (changes requested):
I can't get the add new inventory item to work. I've tried clearing my site data and it's still not showing.

image

src/screens/Inventory/AddInventory.tsx
julianrkung: I just realized what this piece of code is trying to do (always show the `NEW_PRODUCT_LABEL`). Still, It's not working for me for some reason :(

wangannie: Ah I was going off of the code demo in this section https://material-ui.com/components/autocomplete/#creatable because I originally had it in the same format (where you type in the name of the new thing you want to create and it shows up as a new option like `Add "what you typed"`).

src/screens/Inventory/AddInventory.tsx
```diff
@@ -36,23 +47,45 @@ function AddInventory (props: AddInventoryProps) {
...
+      delete product.id; // delete id field to add to Airtable
+      try {
+        productId = await createProduct(product);
+      } catch (error) {
+        console.log('[AddInventory] (createProduct, handleSubmit) Error: ', error);
+        productId = generateOfflineId();
+      }
+      product.id = productId;
+      addProductToRedux(product);
+    }
```
julianrkung: Aren't we putting this in `request.js`? If so, we should be consistent

julianrkung (commented):
Only 1 comment. After it's addressed I'll approve. Good work you're speedy you're so speedy vroom vwoosh skrt skrt

src/screens/Inventory/AddInventory.tsx
wangannie: I moved it into `request.js` and I can update the patch after you check out my (poor) function naming and probably after I merge #66 ?

src/lib/airtable/request.js
```diff
@@ -186,6 +187,21 @@ export const createManyFinancialSummaries = async (records) => {
   return Promise.all(createPromises);
 };
 
+// NONGENERATED: Create a product (inventory type) and add it to Redux
+export const createNewProduct = async (product) => {
```
julianrkung: Can we just modify the `createProduct` function so that we can use the `createProduct` function instead of needing to use `createNewProduct`? Would be confusing to incoming devs.

I was thinking we could modify `createProduct` like the following

```
export const createProduct = async (record) => {
  let productId = '';
  try {
  ...
  productId = await createRecord(Table.Products, record);
  } catch {
  ...
  productId = generateOfflineId()
  }
  ...
  addProductToRedux(...)
  return product.id;
```

so basically what you have except instead of `await createProduct(product)` you do all this logic in `createProduct` and do a `createRecord(Tables.Products, record)` call instead

src/screens/Inventory/AddInventory.tsx
julianrkung: Makes sense, thanks!

julianrkung (approved):
Online had proper Airtable ID and offline produced a proper offline ID.

Looked good when I tried it.

Make sure you regen the patch before merging. Let me know if you want me to check over after you regen

PR #70, opened 2021-04-05
Tiffany/profile frontend

cnnmon:

# What's new in this PR

With this PR, I updated the base customer profile to fit the new Figma changes & added conditions to change the look/data displayed on the profile based on the given customer's meter type. The default is Analog Meter.

## How to review

- Edit a customer's meter type in Airtable (or navigate to the pre-changed customers below)
- Login using testUser
- Navigate to that customer's profile screen

## Screenshots

Analog Meter:
image

Smart Meter:
image

No Meter:
image

Inactive:
image

## Next steps

I put a few TODO's in the code, mostly about linking future pages like the create payments or edit starting meter pages which haven't been made yet. Correct me if I'm wrong!

CC: @julianrkung @micahtyong @wangannie 


package.json
```diff
@@ -34,6 +34,8 @@
     "connected-react-router": "^6.8.0",
     "history": "^4.10.1",
     "i18next": "^19.9.2",
+    "i18next-browser-languagedetector": "^6.1.0",
+    "i18next-http-backend": "^1.2.1",
```
julianrkung: Do we need this? I think this is translation-related stuff. Might be related to Micah's PR but not sure how it got in here

src/components/OutlinedCardList.tsx
```diff
@@ -42,11 +45,17 @@ const useStyles = makeStyles((theme: Theme) =>
       color: theme.palette.text.primary,
       margin: '5px',
     }),
-    fullItems: {
+    fullItems: (props: CardProps) => ({
```
julianrkung: delete the `(props: CardProps)` if unused for `fullItems`

src/screens/Customers/components/TabMenu.tsx
```diff
@@ -26,13 +26,14 @@ const styles = (theme: Theme) =>
...
+  defaultTab: string;
   invoices: MeterReadingRecord[];
   payments: PaymentRecord[];
 }
 
 function TabMenu(props: TabProps) {
-  const { classes, invoices, payments } = props;
-  const [value, setValue] = React.useState('0');
+  const { classes, invoices, payments, defaultTab } = props;
+  const [value, setValue] = React.useState(defaultTab);
```
julianrkung, line 36: can we change the variable name so that it reflects what it does? (Something related to tab)

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -126,35 +157,291 @@ function CustomerProfile(props: CustomerProps) {
...
+      return (
+        <div className={classes.tariffInfo}>
+          {tariffInfo.map((info, index) => (
+            <div key={index} className={classes.tariffInfoItem}>
+              <Typography variant="h3" align={'center'}>
+                {info.number} {info.unit}
+              </Typography>
+              <Typography variant="body1" align={'center'} className={classes.tariffInfoLabel}>{info.label}</Typography>
+            </div>
+          ))}
```
julianrkung: The `if` and `else` returns are basically the same, so we might want to abstract this away into another component? We could call it `TariffPlanInfo` because Annie made an `InventoryInfo` so let's stick w/that convention?

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -126,35 +157,291 @@ function CustomerProfile(props: CustomerProps) {
...
+            { /* Bottom Left */ }
+            <div style={{ float: 'right' }} className={classes.meterInfoContainer}>
+              <div style={{ marginRight: '4px' }} className={classes.meterInfo}>
+                <Typography variant="body1">{meterInfo[1].label}</Typography>
+                <Typography variant="h3">
+                  {meterInfo[1].number} {meterInfo[1].unit}
+                </Typography>
+              </div>
+            </div>
+          </div>
```
julianrkung: Can each of these be made into their own component? (Maybe each component could be one of the `MeterInfoContainer`?). Looks like it's all the same besides what's getting passed into it. This'll help make the code more refactorable when we / the NPO make changes to it

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -126,35 +157,291 @@ function CustomerProfile(props: CustomerProps) {
...
+          <div className={classes.infoContainer}>
+            {getTariffInfo()}
           </div>
-          <div className={classes.section}>
+          <div className={classes.headerWrapper}>
+            <Typography variant="h2">Payment</Typography>
+            <Link className={classes.buttonSecondary}
+              to={{
+              pathname: `${match.url}/records`,
+              state: { invoices: customer.meterReadings, payments: customer.payments, defaultTab: '1' },
```
julianrkung: FYI the `customerData` PR got merged in, so `customer.meterReadings` isn't the way to access the meter readings / payments anymore. The PR modifies `CustomerProfile` and so it shows the new way to grab this info, but lmk if you have trouble after you pull because the conflict is kinda my fault

julianrkung (commented):
The frontend looks great! I think we could cut down on the amount of duplicate code by building out a couple smaller components to use. I some suggestions.

Also, For some reason the meter reading icon wasn't rendering for me? Not sure what was going on there but this person has an analog meter so there should be the lightning bolt icon in the Meter Reading section right?

image

package.json
cnnmon: I'm not sure, I pulled from main.

src/screens/Customers/components/TabMenu.tsx
cnnmon: Wait is default tab not reflective enough? It's just the default number of tab this page should open on.

package.json
micahtyong: Hmm, odd, I thought I removed this. You can run `npm uninstall i18next-browser-languagedetector i18next-http-backend` to get rid of them.

src/screens/Customers/CustomerProfile.tsx
micahtyong: If you do choose to create a component for this, you can also use [map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) to display each JSX element more cleanly! [React's documentation](https://reactjs.org/docs/lists-and-keys.html) has some great examples of this. 

For your case, something like 
```
const listItems = myList.map((position) =>
    // Correct! Key should be specified inside the array.
    <MyCustomTypographyElement key={position.toString()} value={position} />
  );
```
might be useful!

micahtyong: I also noticed that you're hard-coding margin-lefts and margin-rights. This approach is fine for now, but there's also a way to make a 2 by 2 grid without hard-coding! This [StackOverflow](https://stackoverflow.com/questions/24197007/how-can-i-show-three-columns-per-row) post shows how to format six plain old divs into a grid with two rows and three columns.

src/components/OutlinedCardList.tsx
```diff
@@ -67,11 +76,11 @@ export default function OutlinedCardList(props: CardProps): JSX.Element {
       <div key={key} className={ classes.itemWrapper }>
         <div className={rightIcon ? classes.items : classes.fullItems}>
           <Typography variant="body1" align={columns ? 'center' : 'inherit'}>{label}</Typography>
-          <Typography variant="h3" align={columns ? 'center' : 'inherit'} color={primary ? 'primary' : 'inherit'}>
+          <Typography variant="h3" align={columns ? 'center' : 'inherit'} className={primary ? classes.red : props.textColor}>
```
micahtyong: Nice use of ternaries!

src/screens/Customers/CustomerProfile.tsx
micahtyong: Even better would just to use another ternary on `className`, since that's the only thing you're conditioning on. So something like 
```
<Typography variant="h3" align={'center'} className={customer.meterType == "Inactive" ? classes.gray : ''}> 
```

src/screens/Customers/CustomerProfile.tsx
cnnmon: Would it be ok if I made a MeterInfo component to contain all 4 conditions? I think there's a lot of subtle style changes between the meter types (ex. grayed text, links, icons) that'd be hard to generalize.

cnnmon: Actually wait lemme try this first

src/screens/Customers/components/TabMenu.tsx
julianrkung: Oh sorry I meant `value` and `setValue` weren't descriptive names, but I noticed that they were there before. `defaultTab` is a good name! I guess there's no need to change then if you don't want since it's out of scope of PR!

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -63,62 +68,42 @@ function CustomerProfile(props: CustomerProps) {
...
+  let meterInfo: CardPropsInfo[] = [
+    { number: startingReading? startingReading.amountBilled.toString() : '0', label: 'Starting Meter', unit: 'kWh' },
     { number: periodUsage.toString(), label: 'Period Usage', unit: 'kWh' },
+    { number: currReading ? currReading.toString() : '0', label: 'Ending Meter', unit: 'kWh' },
     { number: amountBilled.toString(), label: 'Amount Billed', unit: 'kS' },
   ];
-  const balanceInfo: CardPropsInfo[] = [{ number: customer.outstandingBalance.toString(), label: 'Remaining Balance', unit: 'kS' }];
-  const readingInfo: CardPropsInfo[] = [{ number: currReading ? currReading.amountBilled.toString() : UNDEFINED_AMOUNT, label: 'Current Reading', unit: 'kWh' }];
+  let balanceInfo: CardPropsInfo[] = [{ number: customer.outstandingBalance.toString(), label: 'Remaining Balance', unit: 'kS' }];
+  let readingInfo: CardPropsInfo[] = [{ number: currReading? currReading.amountBilled.toString() : '0', label: 'Last Recorded Reading', unit: 'kWh' }];
```
julianrkung: Should show the current reading in kWh not the amount billed for that reading

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -63,62 +68,42 @@ function CustomerProfile(props: CustomerProps) {
...
-  const minUnits = customerTariff ? customerTariff?.minUnits : UNDEFINED_AMOUNT;
-
-  const tariffInfo: CardPropsInfo[] = [
-    { number: fixedTariff.toString(), label: 'Fixed Tariff', unit: 'MMK' },
-    { number: unitTariff.toString(), label: 'Unit Tariff', unit: 'MMK' },
-    { number: minUnits.toString(), label: 'Minimum Units Used', unit: '' },
-  ]
+  const fixedTariff = customerTariff ? customerTariff?.fixedTariff : '0';
+  const unitTariff = customerTariff ? customerTariff?.tariffByUnit : '0';
+  const minUnits = customerTariff ? customerTariff?.minUnits : '0';
```
julianrkung: You should pull from main. This is now called `freeUnits` in both the figma and on `main`

src/screens/Customers/components/TariffPlanInfo.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { createStyles, withStyles, Theme } from '@material-ui/core/styles';
+import { Typography } from '@material-ui/core';
+import { CardPropsInfo } from '../../../components/OutlinedCardList';
+
+const styles = (theme: Theme) =>
+  createStyles({
+    root: {
+      marginTop: '10px',
+      border: '1px solid',
+      borderColor: theme.palette.divider,
```
julianrkung: Figma was recently updated to have a different border color. Can you update here also?

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -63,62 +68,42 @@ function CustomerProfile(props: CustomerProps) {
...
-  const amountBilled: number | string = currReading ? getAmountBilled(currReading) : UNDEFINED_AMOUNT;
+  const periodUsage: number | string = currReading ? getPeriodUsage(currReading, startingReading) : '0';
+  const amountBilled: number | string = currReading ? getAmountBilled(currReading) : '0';
 
-  const meterInfo: CardPropsInfo[] = [
-    { number: startingReading ? startingReading.amountBilled.toString() : UNDEFINED_AMOUNT, label: 'Starting Meter', unit: 'kWh' },
+  let meterInfo: CardPropsInfo[] = [
+    { number: startingReading? startingReading.amountBilled.toString() : '0', label: 'Starting Meter', unit: 'kWh' },
     { number: periodUsage.toString(), label: 'Period Usage', unit: 'kWh' },
+    { number: currReading ? currReading.toString() : '0', label: 'Ending Meter', unit: 'kWh' },
```
julianrkung: Should this be `currReading.reading`? `currReading` is a `MeterReadingRecord`

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -63,62 +68,42 @@ function CustomerProfile(props: CustomerProps) {
...
   const startingReading: MeterReadingRecord | undefined = getStartingReading(customer);
-  const periodUsage: number | string = currReading ? getPeriodUsage(currReading, startingReading) : UNDEFINED_AMOUNT;
-  const amountBilled: number | string = currReading ? getAmountBilled(currReading) : UNDEFINED_AMOUNT;
+  const periodUsage: number | string = currReading ? getPeriodUsage(currReading, startingReading) : '0';
+  const amountBilled: number | string = currReading ? getAmountBilled(currReading) : '0';
 
-  const meterInfo: CardPropsInfo[] = [
-    { number: startingReading ? startingReading.amountBilled.toString() : UNDEFINED_AMOUNT, label: 'Starting Meter', unit: 'kWh' },
+  let meterInfo: CardPropsInfo[] = [
+    { number: startingReading? startingReading.amountBilled.toString() : '0', label: 'Starting Meter', unit: 'kWh' },
```
julianrkung, line 85: This needs to be changed to a meter reading instead of a billing amount, but the actual reading it's supposed isn't in the schema/interface yet. It's eventually supposed to be `currentCustomer.startingMeterReading`, but I can add that on the Meter Reading PR (or whenever after this PR is merged because that info isn't available yet). Can you put a TODO for that?

src/screens/Customers/components/MeterInfoContainer.tsx
```diff
@@ -0,0 +1,86 @@
...
+  if (editable) {
+    return (
+      <div className={classes.root} style={{ float: floatRight ? 'right' : 'left' }}>
+        <div className={classes.info} style={{
+          marginRight: floatRight ? undefined : '4px',
+          marginLeft: floatRight ? '4px' : undefined,
+          paddingTop: '0px',
+          backgroundColor: 'white' }}
+        >
+          <div className={classes.inline}>
```
julianrkung: Could we have the styling dependent on the `editable` prop

and then conditionally render the `<div className={classes.inline}> ... </div>`? That way we won't need the `if / else` with duplicate code. As far as I can tell, that's the only difference between the 2. Would look something like this. 

```
<div className={classes.root} style={{ float: floatRight ? 'right' : 'left' }}>
        <div className={classes.info} style={{
          marginRight: floatRight ? undefined : '4px',
          marginLeft: floatRight ? '4px' : undefined,
          paddingTop: editable ? '0px' : (whatever pixels it is),
          backgroundColor: editable ? 'white' : 'inherit' }}
        >
            <div className={classes.inline}>
                <Typography ... > {info.label} <Typography />
                {editable ? renderEditableIcon() : null}
            </div>
          <Typography variant="h3" style={{ color: gray ? 'rgba(189,189,189,1)' : undefined }}>
            {info.number} {info.unit}
          </Typography>
       </div>
</div>
```

Also, is there a reason that the styles are inlined here instead of as a part of `classes.info`? You could do something similar as what you did in [OutlinedCardList](https://github.com/calblueprint/meepanyar/pull/70/files#diff-8ff1758d7a7916261b5e069894929d0fce367a367359a31926497cf2123921c5R27) where the styling is dependent on the props passed (in this case `editable`). I think as long as this `if/else` statement is shortened so that the code isn't repeated, I'm fine with it though!

src/screens/Customers/components/TariffPlanInfo.tsx
```diff
@@ -0,0 +1,90 @@
...
+            </div>
+          ))}
+        </div>
+      </div>
+    );
+  } else {
+    const tariffInfo : CardPropsInfo[] = [
+      { number: fixedTariff , label: 'Fixed\nTariff', unit: 'Ks' },
+      { number: unitTariff, label: 'Unit\nTariff', unit: 'Ks' },
+      { number: minUnits, label: 'Minimum\nUnits', unit: 'kWh' },
```
julianrkung: Now is `freeUnits` (also in the interface) sorry for the changeup ;-;

src/screens/Customers/components/TariffPlanInfo.tsx
```diff
@@ -0,0 +1,90 @@
...
+            <div key={index} className={classes.item}>
+              <Typography variant="h3" align={'center'} style={{ color: 'rgba(189,189,189,1)' }}>
+                {info.number} {info.unit}
+              </Typography>
+              <Typography variant="body1" align={'center'} className={classes.label}>{info.label}</Typography>
+            </div>
+          ))}
+        </div>
+      </div>
+    );
```
julianrkung: Is the only difference between the 2 returns the color of the Typography? If so, you can do an `if else` to set the value of the `tariffInfo` and then use `className.gray` on the typography if the `meterType === 'Inactive`. That way, the code is more maintainable since there's only 1 copy of it (no need to have 2 return statements)

src/screens/Customers/components/TariffPlanInfo.tsx
```diff
@@ -0,0 +1,90 @@
...
+  meterType: string;
+  fixedTariff: string;
+  unitTariff: string;
+  minUnits: string;
+}
+
+function TariffPlanInfo(props: TariffPlanInfoProps) {
+  const { classes, undefinedAmount, meterType, fixedTariff, unitTariff, minUnits } = props;
+
+  if (meterType === "Inactive") {
```
julianrkung: Also, I don't know if we need a `meterType === 'Inactive'` because that would just mean the customer is inactive right? We can swap this out later, but just wanted to know if there's a distinction that I don't know about

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -133,35 +118,198 @@ function CustomerProfile(props: CustomerProps) {
...
+            { /* Bottom Right */ }
+            <MeterInfoContainer
+              floatRight={true}
+              info={meterInfo[3]}
+            />
+            { /* Bottom Left */ }
+            <MeterInfoContainer
+              floatRight={false}
+              info={meterInfo[1]}
+            />
```
julianrkung: Can the `if` statements just determine what to pass to the `MeterInfoContainer`s and just return 1 set of `MeterInfoContainers` instead of having 4 returns? I was thinking something along the lines of 

```
let meterInfo;
let readingInfo;
let topLeftGray;
let topRightGray;
let bottomRightGray;
let bottomLeftGray;

if (customer.meterType === 'Analog Meter') {
    meterInfo = .... 
    readingInfo = ...
    ...
} else if (customer.meterType === 'SmartMeter) {
    meterInfo = ...
    readingInfo = ...
    ...
} else {
    ...
}

return (
        <div>
          <OutlinedCardList info={readingInfo} primary={false} grayBackground={true} />
          <div className={classes.section}>
            { /* Top Left */ }
            <MeterInfoContainer
              floatRight={false}
              info={meterInfo[0]}
              gray={topLeftGray}
            />
            { /* Top Right */ }
            <MeterInfoContainer
              floatRight={true}
              info={meterInfo[2]}
              gray={topRightGray}
            />
            { /* Bottom Right */ }
            <MeterInfoContainer
              floatRight={true}
              info={meterInfo[3]}
              gray={bottomRightGray}
            />
            { /* Bottom Left */ }
            <MeterInfoContainer
              floatRight={false}
              info={meterInfo[1]}
              gray={bottomLeftGray}
            />
          </div>
        </div>
)

```

Cuts the amount of code by a lot and is easier to read and easier to maintain since there's only 1 return

julianrkung (commented):
I took a look and left some comments. Matches Figma really well, just needs some cleanup from what I can tell, great job! :)

src/screens/Customers/components/TariffPlanInfo.tsx
cnnmon: oooh true ill add a Todo there

src/lib/airtable/interface.ts
```diff
@@ -71,6 +71,7 @@ export interface CustomerRecord {
   customerUpdates: CustomerUpdateRecord[];
   totalAmountBilledfromInvoices: number;
   totalAmountPaidfromPayments: number;
+  meterType: string;
```
wangannie: Might make sense to use an enum here instead (see `PurchaseRequestStatus` in `inventoryDataSlice`)

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -139,29 +115,127 @@ function CustomerProfile(props: CustomerProps) {
     );
   };
 
+  const getPaymentInfo = () => {
+    if (customer.meterType === "Inactive") {
+      balanceInfo = [{ number: UNDEFINED_AMOUNT, label: 'Remaining Balance', unit: '' }];
+      return (
+        <OutlinedCardList info={balanceInfo} primary={false} grayBackground={true} grayText={true} />
```
wangannie: When you have these `boolean` props, you don't need to do `prop={true}` -- you can just include the prop name
`<OutlinedCardList ... grayBackground grayText />`

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -139,29 +115,127 @@ function CustomerProfile(props: CustomerProps) {
...
+      ];
+    }
+
+    return (
+      <div>
+        <OutlinedCardList
+          info={readingInfo}
+          primary={false}
+          rightIcon={readingCardState === 0 ? getAddButton('/meter-readings/create') : undefined}
+          grayBackground={readingCardState === 1 || readingCardState === 2 ? true : undefined} grayText={readingCardState === 2 ? true : undefined}
```
wangannie: For these booleans, you don't need ` ? true : undefined` -- you can just do `grayText={readingCardState === 2` since that evaluates to true/false anyways

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -139,29 +115,127 @@ function CustomerProfile(props: CustomerProps) {
...
+            unitTariff={unitTariff.toString()}
+            freeUnits={freeUnits.toString()}
+          />
+          <div className={classes.headerWrapper}>
+            <Typography variant="h2">Payment</Typography>
+            <Link className={classes.buttonSecondary}
+              to={{
+              pathname: `${match.url}/records`,
+              state: { invoices: meterReadings, payments: payments, defaultTab: '1' },
+            }}>View All</Link>
```
wangannie: Once you merge in main with the changes from #80, update this (and all other text buttons) to use the text variant of the button base component.

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -139,29 +115,127 @@ function CustomerProfile(props: CustomerProps) {
...
             <Typography variant="h2">
               Meter Reading
             </Typography>
-            <OutlinedCardList info={readingInfo} primary={true} rightIcon={getAddButton()} />
-            <OutlinedCardList info={meterInfo} primary={false} />
+            <Link className={classes.buttonSecondary}
+              to={{
+              pathname: `${match.url}/records`,
+              state: { invoices: meterReadings, payments: payments, defaultTab: '0' },
+            }}>View All</Link>
```
wangannie: Same as above -- replace with base component

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -79,56 +85,26 @@ function CustomerProfile(props: CustomerProps) {
...
-        startIcon={<ListAltOutlinedIcon />}
-        disableElevation={true}
-      >
-        Records
-      </Button>
-    );
-  };
-
-  const getAddButton = () => {
+  const getAddButton = (path: String) => {
```
wangannie: Switch to lowercase `string`

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -79,56 +85,26 @@ function CustomerProfile(props: CustomerProps) {
...
     return (
       <div className={classes.buttonPrimary}>
         <IconButton
           component={Link}
           to={{
-            pathname: `${match.url}/meter-readings/create`,
+            pathname: `${match.url}${path}`,
             state: { invoices: meterReadings, payments: payments },
           }}
           size="small"
```
wangannie: I realized this should definitely be extracted out into a base component eventually. Can you add a TODO @wangannie  for that?

src/components/OutlinedCardList.tsx
```diff
@@ -47,6 +50,12 @@ const useStyles = makeStyles((theme: Theme) =>
       color: theme.palette.text.primary,
       margin: '5px',
     },
+    red: {
+      color: theme.palette.info.main,
+    },
+    text: (props: CardProps) => ({
+      color: props.grayText ? 'rgba(189,189,189,1)' : theme.palette.text.primary,
```
wangannie: In general, we shouldn't be introducing rgb's when we've been using hex codes everywhere else. But ALSO, since this is equivalent to #bdbdbd, use the palette color that corresponds to that (`text.disabled` once you pull from main)

src/components/OutlinedCardList.tsx
```diff
@@ -67,11 +76,11 @@ export default function OutlinedCardList(props: CardProps): JSX.Element {
       <div key={key} className={ classes.itemWrapper }>
         <div className={rightIcon ? classes.items : classes.fullItems}>
           <Typography variant="body1" align={columns ? 'center' : 'inherit'}>{label}</Typography>
-          <Typography variant="h3" align={columns ? 'center' : 'inherit'} color={primary ? 'primary' : 'inherit'}>
+          <Typography variant="h3" align={columns ? 'center' : 'inherit'} className={primary ? classes.red : classes.text}>
```
wangannie: I'd change this back to `color=` since the class is literally just a color anyways.

Also, passing in `primary` to indicate that the text should be red (at least what I think is happening?) is very confusing -- it'd make more sense to just pass in a color string prop here so you don't have to use this nested ternary and you could just do something like `<Typography ... color={props.color || 'inherit' }>`

I believe it'll give you issues trying to pass a string in for this, so in the params you'll probably have to specify it like 
`color?: "primary" | "error" | "textPrimary" |... (whatever material palette colors are relevant)`

I know that might've been confusing so feel free to follow up/dm me if you have questions about this.

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -36,17 +33,26 @@ const styles = (theme: Theme) =>
...
     buttonSecondary: {
-      borderRadius: '12px',
       color: theme.palette.primary.main,
-      backgroundColor: theme.palette.secondary.main,
+      fontSize: '12px',
+      textTransform: 'uppercase',
+      letterSpacing: '0.10em',
+      '&:hover': {
+        textDecoration: 'none',
+        color: theme.palette.primary.main,
```
wangannie: `buttonSecondary` should be removed (see my other comments on using the text button base component)

but also why do we have hover states on a mobile app 😅  be careful with what you copy from the internet

src/components/OutlinedCardList.tsx
wangannie: Also, once you merge in `main`, I've replaced the `info` color with `error`, so pass in "error" to get this red.

src/screens/Customers/components/MeterInfoContainer.tsx
```diff
@@ -0,0 +1,72 @@
...
+    </IconButton>
+  );
+
+  return (
+    <div className={classes.root} style={{ float: floatRight ? 'right' : 'left' }}>
+        <div className={classes.info} style={{
+          marginRight: floatRight ? undefined : '4px',
+          marginLeft: floatRight ? '4px' : undefined,
+          borderColor: editable ? '#BDBDBD' : '#F7F9FC',
+          backgroundColor: editable ? 'white' : '#F7F9FC' }}
```
wangannie: You should really never be using custom hex codes -- use the corresponding colors in the material theme palette whenever possible. 

Also, make sure you pull from main because some colors were adjusted in my latest PR

src/screens/Customers/components/MeterInfoContainer.tsx
```diff
@@ -0,0 +1,72 @@
...
+        >
+          <div className={classes.inline}>
+            <Typography variant="body1">{info.label}</Typography>
+            {editable ? renderEditableIcon() : null}
+          </div>
+          <Typography variant="h3" style={{ color: gray ? 'rgba(189,189,189,1)' : undefined }}>
```
wangannie: Don't use RGBA! Replace with material palette color

src/screens/Customers/components/MeterInfoContainer.tsx
```diff
@@ -0,0 +1,72 @@
...
+  floatRight: boolean;
+  info: CardPropsInfo;
+  editable?: boolean;
+  gray?: boolean;
+}
+
+function MeterInfoContainer(props: MeterInfoContainerProps) {
+  const { classes, floatRight, info, editable, gray } = props;
+
+  const renderEditableIcon = () => (
```
wangannie: It seems like in most other places we use `get___Button` instead of `renderIcon`. Rename this to something like `getEditIconButton` to match that pattern

src/screens/Customers/components/MeterInfoContainer.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+interface MeterInfoContainerProps {
+  classes: { root: string; info: string; icon: string; inline: string; };
+  floatRight: boolean;
+  info: CardPropsInfo;
+  editable?: boolean;
+  gray?: boolean;
+}
+
+function MeterInfoContainer(props: MeterInfoContainerProps) {
```
wangannie: I'm confused why this had to be a separate component when it basically seems like `OutlinedCardList` but with slightly different styling + an edit button. But for the sake of time, maybe just leave a TODO comment to consolidate the two?

cc @julianrkung

src/screens/Customers/components/MeterInfoContainer.tsx
```diff
@@ -0,0 +1,72 @@
...
+      color: theme.palette.primary.main,
+    },
+    inline: {
+      display: 'inline-flex',
+    },
+  });
+
+interface MeterInfoContainerProps {
+  classes: { root: string; info: string; icon: string; inline: string; };
+  floatRight: boolean;
```
wangannie: This would be a lot cleaner if you used flexbox! Then you wouldn't need a  `floatRight` or custom margins because you'd just wrap these cards in a flex container to get the grid layout.

src/screens/Customers/components/MeterInfoContainer.tsx
```diff
@@ -0,0 +1,72 @@
...
+      width: '50%',
+      display: 'flex',
+    },
+    info: {
+      width: '100%',
+      marginBottom: '8px',
+      padding: '0px 12px 4px',
+      border: '1px solid',
+      height: '60px',
+      maxHeight: '60px',
```
wangannie: Why do you have height and maxHeight?

wangannie: Also in general, avoid setting set pixel heights, aside from `minHeight`. If I shrink my screen width, you can see how `OutlinedCardList` adapts the height whereas this overflows.

image

src/screens/Customers/components/MeterInfoContainer.tsx
```diff
@@ -0,0 +1,72 @@
...
+      display: 'inline-flex',
+    },
+  });
+
+interface MeterInfoContainerProps {
+  classes: { root: string; info: string; icon: string; inline: string; };
+  floatRight: boolean;
+  info: CardPropsInfo;
+  editable?: boolean;
+  gray?: boolean;
```
wangannie: It seems like this is only ever either editable or gray, so could you eliminate the `gray` prop?

src/screens/Customers/components/TariffPlanInfo.tsx
```diff
@@ -0,0 +1,76 @@
...
+      { number: freeUnits, label: 'Free\nUnits', unit: 'kWh' },
+    ];
+  }
+
+  return (
+    <div className={classes.root} style={{ backgroundColor: meterType === "Inactive" ? '#F7F9FC' : 'inherit', borderColor: meterType === "Inactive" ? '#F7F9FC' : 'inherit' }}>
+      <div className={classes.content}>
+        {tariffInfo.map((info, index) => (
+          <div key={index} className={classes.item}>
+            <Typography variant="h3" align={'center'} style={{color: meterType === "Inactive" ? 'rgba(189,189,189,1)' : undefined}}>
```
wangannie: No RGBA!

src/screens/Customers/components/TariffPlanInfo.tsx
```diff
@@ -0,0 +1,76 @@
...
+  } else {
+    tariffInfo = [
+      { number: fixedTariff , label: 'Fixed\nTariff', unit: 'Ks' },
+      { number: unitTariff, label: 'Unit\nTariff', unit: 'Ks' },
+      { number: freeUnits, label: 'Free\nUnits', unit: 'kWh' },
+    ];
+  }
+
+  return (
+    <div className={classes.root} style={{ backgroundColor: meterType === "Inactive" ? '#F7F9FC' : 'inherit', borderColor: meterType === "Inactive" ? '#F7F9FC' : 'inherit' }}>
```
wangannie: Replace HEX with palette

src/screens/Customers/components/TariffPlanInfo.tsx
```diff
@@ -0,0 +1,76 @@
...
+    ];
+  } else {
+    tariffInfo = [
+      { number: fixedTariff , label: 'Fixed\nTariff', unit: 'Ks' },
+      { number: unitTariff, label: 'Unit\nTariff', unit: 'Ks' },
+      { number: freeUnits, label: 'Free\nUnits', unit: 'kWh' },
+    ];
+  }
+
+  return (
```
wangannie: I'm not sure why you re-wrote the tariff info when it was originally just one line:
`<OutlinedCardList info={tariffInfo} primary={false} columns />` 
and it appears that the only difference is that the text is now on the bottom (which could easily be implemented in `OutlinedCardList` by passing in a prop). I understand why it's clean to have this as a dedicated component for the `tariffInfo` logic, but I think it'd make sense if you replace this with the original one-liner.

src/components/OutlinedCardList.tsx
```diff
@@ -13,16 +13,19 @@ interface CardProps {
   primary: boolean;
   rightIcon?: JSX.Element;
   columns?: boolean;
+  grayBackground?: boolean;
+  grayText?: boolean;
```
wangannie: if you make the change I described below, you shouldn't need this `greyText`

src/screens/Customers/components/MeterInfoContainer.tsx
wangannie: jk I saw in the designs, there is a version of `OutlinedCardList` with an edit button?? Do we know if this is an actual use case?
Screen Shot 2021-04-09 at 1 20 45 PM

src/components/OutlinedCardList.tsx
```diff
@@ -13,16 +13,19 @@ interface CardProps {
   primary: boolean;
   rightIcon?: JSX.Element;
   columns?: boolean;
+  grayBackground?: boolean;
```
wangannie: If I'm understanding this correctly, it seems like grey background represents a 'read only' state, so I think it would make more sense to use a prop like `readOnly` that represents its actual use case, not just the actual color.

wangannie (changes requested):
Feel free to DM me if you have any questions!

src/screens/Customers/components/MeterInfoContainer.tsx
julianrkung: I don't know the case for editing a payment (actually, we might not want to make payments editable on the app. I'll think more about it and ask Ashley). The use case for editing a meter reading is if they inputted it wrong (since there's only supposed to be 1 meter reading per period) and need to edit it

src/components/OutlinedCardList.tsx
cnnmon: if you look on the figma under smart meter vs no meter/inactive, there is a distinction between when there's gray text and gray background

src/screens/Customers/components/MeterInfoContainer.tsx
cnnmon: this is also a case of if you look at smart meter vs no meter, there's not always gray text AND gray background

src/screens/Customers/components/MeterInfoContainer.tsx
cnnmon: to make sure the grid format stayed the same. also would people really view the app at such a small screen width?

src/components/OutlinedCardList.tsx
cnnmon: see comment above ^ if there's a distinction between gray background & text i think i want to keep the distinction in the naming

src/lib/redux/customerDataSlice.ts
```diff
@@ -16,6 +16,13 @@ const meterReadingsAdapter = createEntityAdapter<MeterReadingRecord>({
     sortComparer: (a, b) => moment(b.date).diff(a.date),
 });
 
+export enum MeterType {
```
julianrkung, line 19: Nice :) enums will definitely save future devs from headaches

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -63,72 +60,43 @@ function CustomerProfile(props: CustomerProps) {
...
-  const getAddButton = () => {
+  const getAddButton = (path: string) => {
+    //TODO: separate into base component @wangannie
     return (
       <div className={classes.buttonPrimary}>
         <IconButton
           component={Link}
           to={{
-            pathname: `${match.url}/meter-readings/create`,
+            pathname: `${match.url}${path}`,
```
julianrkung: Nice change for reusability! Kinda nit-picky, but it might be good to instead do `${match.url}/${path}` because there should always be the slash (make sure to change `getAddButton('/meter-reading/create')` to `getAddButton('meter-reading/create')` if you do this change)

src/components/OutlinedCardList.tsx
```diff
@@ -65,18 +75,27 @@ export default function OutlinedCardList(props: CardProps): JSX.Element {
...
-          }
-        </div>
+            {/* Split unit to new line if in column layout */}
+            {props.columns &&
+              <Typography variant="h3" align={columns ? 'center' : 'inherit'} color={primary ? 'primary' : 'inherit'}>
+                {unit}
+              </Typography>
+            }
+          </div>
+        }
```
julianrkung: Is there any way to break this into a single `return` instead of doing this ternary (prob via abstracting portions of the return into helper functions)?

As a general rule, you don't want to do large conditional returns where there's a lot of overlap in the code. It quickly becomes hard to read and see what the difference are (and it's more code to maintain when people want to come in and make changes since they need to make changes to both the conditionals in the ternary most of the time)


I came up with the following really quickly, but the naming of functions isn't that great (and doesn't deal with the interaction between `props.columns` and `props.reverse` with the unit supposedly being rendered on a new line sometimes (see comment above)

```
  const getLabeledNumber = (
    key: number,
    number: string,
    label: string,
    unit: string,
    primary: boolean,
    rightIcon: JSX.Element | null,
    columns: boolean,
  ) => {

    const getFormattedLabel = () => (<Typography align={props.columns ? 'center' : 'inherit'} className={props.reverse ? classes.reverseLabel : undefined}>{label}</Typography>);
    const getFormattedNumber = () => (<Typography
      variant="h3"
      align={columns ? 'center' : 'inherit'}
      color={primary ? 'error' : 'inherit'}
      className={props.grayText ? classes.grayText : undefined}
    >
      {/* In column layout, we split unit into its own line */}
      {number} {unit}
    </Typography>)

    return (
      <div key={key} className={classes.itemWrapper}>
        <div className={rightIcon ? classes.items : classes.fullItems}>
          {props.reverse ? getFormattedNumber() : getFormattedLabel()}
          {props.reverse ? getFormattedLabel() : getFormattedNumber()}
        </div>
        {rightIcon}
      </div>
    )

```

src/components/OutlinedCardList.tsx
```diff
@@ -65,18 +75,27 @@ export default function OutlinedCardList(props: CardProps): JSX.Element {
...
+          :
+          <div className={rightIcon ? classes.items : classes.fullItems}>
+            <Typography variant="body1" align={columns ? 'center' : 'inherit'}>{label}</Typography>
+            <Typography variant="h3" align={columns ? 'center' : 'inherit'} color={primary ? 'error' : 'inherit'} className={props.grayText ? classes.grayText : undefined}>
+              {number} {props.columns ? '' : unit}
             </Typography>
-          }
-        </div>
+            {/* Split unit to new line if in column layout */}
+            {props.columns &&
```
julianrkung: `props.columns` is the same as `columns` here right? We should maintain the same convention if so (sidenote: I also don't see the need to pass `columns` in as a parameter to `getLabeledNumber`, since it's always the value of `props.columns` anyways, and the same for `rightIcon`. You could probably delete the `columns` and `rightIcon` parameters from `getLabeledNumber` and do `const { columns, rightIcon } = props` at the top of the `OutlinedCardList` function and it'd be cleaner

src/components/OutlinedCardList.tsx
```diff
@@ -65,18 +75,27 @@ export default function OutlinedCardList(props: CardProps): JSX.Element {
...
+        {props.reverse ?
+          <div className={rightIcon ? classes.items : classes.fullItems}>
+            <Typography variant="h3" align={columns ? 'center' : 'inherit'} color={primary ? 'error' : 'inherit'} className={props.grayText ? classes.grayText : undefined}>
+              {number} {unit}
+            </Typography>
+            <Typography variant="body1" align={columns ? 'center' : 'inherit'} className={classes.reverseLabel}>{label}</Typography>
+          </div>
+          :
+          <div className={rightIcon ? classes.items : classes.fullItems}>
+            <Typography variant="body1" align={columns ? 'center' : 'inherit'}>{label}</Typography>
```
julianrkung: NIT: Typography variant is default to `body1` so no need to specify

source: [here](https://material-ui.com/api/typography/)

src/components/OutlinedCardList.tsx
```diff
@@ -65,18 +75,27 @@ export default function OutlinedCardList(props: CardProps): JSX.Element {
...
-          {/* Split unit to new line if in column layout */}
-          {props.columns && 
-            <Typography variant="h3" align={columns ? 'center' : 'inherit'} color={primary ? 'primary' : 'inherit'}>
-              {unit}
+        {props.reverse ?
+          <div className={rightIcon ? classes.items : classes.fullItems}>
+            <Typography variant="h3" align={columns ? 'center' : 'inherit'} color={primary ? 'error' : 'inherit'} className={props.grayText ? classes.grayText : undefined}>
+              {number} {unit}
+            </Typography>
+            <Typography variant="body1" align={columns ? 'center' : 'inherit'} className={classes.reverseLabel}>{label}</Typography>
```
julianrkung: NIT: Typography variant is default to `body1` so no need to specify

source: [here](https://material-ui.com/api/typography/)

src/lib/redux/customerDataSlice.ts
julianrkung: Also as a future note, this should've prob be in `interface.ts` but since `PurchaseRequestStatus` is in `inventoryDataSlice` we can keep it here for consistency. More my bad for not nit-picking where `PurchaseRequestStatus` was when it was initially merged

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -139,29 +107,161 @@ function CustomerProfile(props: CustomerProps) {
...
+      tariffInfoGray = true;
+    }
+    return (
+      <OutlinedCardList info={tariffInfo} primary={false} columns grayBackground={tariffInfoGray} grayText={tariffInfoGray} reverse />
+    );
+  }
+
   return (
-    <BaseScreen leftIcon="backNav" title={customer.name} rightIcon="edit" match={match}>
+    <BaseScreen leftIcon="backNav" title={`${customer.meterNumber}, ${customer.name}`} rightIcon="edit" match={match}>
```
julianrkung: Should be `customer.customerNumber` but we don't have that column populated in our Customers table (very sad). Still, I think we should use that here and add `customerNumber` to `interface.ts`

image

src/screens/Customers/components/MeterInfoContainer.tsx
julianrkung: +1 on this. I also think it should be done this way for responsiveness. @wangannie can help if you have trouble with this conversion but it's best to use flexbox here.

I also can't tell the difference between this component vs `OutlinedCardList` besides the gray (and the icon on `Starting Meter`) on the Figma. @cnnmon if we fixed the `grayBackground / grayText` issue, would you be down to use `OutlinedCardList` for this instead? If you think it deserve its own component, I'm down with that too but just wanted to know rationale (though I think at least we could take the styling from `OutlinedCardList` to make it responsive and prevent https://github.com/calblueprint/meepanyar/pull/70/files#r610876674)

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -139,29 +107,161 @@ function CustomerProfile(props: CustomerProps) {
...
+    } else if (customer.meterType === MeterType.NO_METER) {
+      meterGrayText = true;
+      meterGrayBackground = true;
+      topLeftGrayBackground = true;
+      topRightGrayBackground = true;
+      bottomRightGrayBackground = true;
+      bottomLeftGrayBackground = true;
+      topLeftGrayText = true;
+      topRightGrayText = true;
+      bottomRightGrayText = true;
```
julianrkung: If we can just use black text for all logic we could simplify this a lot. Having the conversation with Ashley here
https://calblueprint.slack.com/archives/C01S01Z7VUN/p1618163230039600

src/components/OutlinedCardList.tsx
```diff
@@ -65,18 +75,27 @@ export default function OutlinedCardList(props: CardProps): JSX.Element {
...
+            <Typography variant="h3" align={columns ? 'center' : 'inherit'} color={primary ? 'error' : 'inherit'} className={props.grayText ? classes.grayText : undefined}>
+              {number} {props.columns ? '' : unit}
             </Typography>
-          }
-        </div>
+            {/* Split unit to new line if in column layout */}
+            {props.columns &&
+              <Typography variant="h3" align={columns ? 'center' : 'inherit'} color={primary ? 'primary' : 'inherit'}>
+                {unit}
+              </Typography>
```
julianrkung: The relationship between `props.columns` and `props.reverse` is unclear because it looks like we render the units in another line if `props.reverse = false and props.columns=true` while we render the units on the same line if `props.reverse=true and props.columns=true` which doesn't seem like intuitive behavior.

The good news is that it looks like rendering TariffPlans is the only instance where we use `columns` (I only did a quick search though, might be wrong). So we can change without worrying about breaking other parts of app

julianrkung (commented):
Left comments. Didn't look too closely at the CSS

src/components/OutlinedCardList.tsx
cnnmon: Wait so I'm confused, what am I supposed to do with this section?

julianrkung: I think https://github.com/calblueprint/meepanyar/pull/70#discussion_r611241208 fixes the issue, though we lose the fact that we "Split unit to new line if in column layout", which I don't know if that's important. I think this comment was more to show what might've been a bug / confusing code here

src/components/OutlinedCardList.tsx
```diff
@@ -13,65 +16,99 @@ interface CardProps {
...
+    reverseLabel: {
+      wordSpacing: '100vw',
+      lineHeight: '1.2',
+    },
+    editButton: {
+      float: 'right',
+    },
+    editIcon: {
+      color: theme.palette.primary.main,
+      fontSize: '14px',
```
julianrkung, line 64: image

How should overflow be dealt with here? Overflowing of kWh makes one of the boxes bigger. Should all box heights / widths be set to account for overflow.

Please advise @wangannie

src/components/OutlinedCardList.tsx
```diff
@@ -13,65 +16,99 @@ interface CardProps {
...
 
-
 export default function OutlinedCardList(props: CardProps): JSX.Element {
   const classes = useStyles(props);
 
+  const getEditButton = (path: string) => {
+    return (
+      <IconButton
+        component={Link}
+        to={{ pathname: path }}
```
julianrkung: NIT: i think you can just put to={path} here

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -63,105 +62,183 @@ function CustomerProfile(props: CustomerProps) {
...
+  const getReadingInfo = () => {
+    let meterReadOnly = true;
+    let topLeftReadOnly = true;
+    let topRightReadOnly = true;
+    let bottomRightReadOnly = true;
+    let bottomLeftReadOnly = true;
+
+    if (customer.meterType === MeterType.ANALOG_METER) {
+      meterReadOnly = false;
+      topLeftReadOnly = false;
```
julianrkung: Wait only 2 of these have the possibility of not being readOnly, so you just need to have variables for those 2, the others can just be set in the return as `true`

julianrkung (commented):
Left comment to clean up `getReadingInfo` We can approve with @wangannie OK / advice

src/components/OutlinedCardList.tsx
wangannie: Responded via Slack but this is a design question for Ashley.

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -63,105 +62,180 @@ function CustomerProfile(props: CustomerProps) {
...
+      </IconButton>
     );
   };
 
-  const getAddButton = () => {
+  const getPaymentInfo = () => {
+    if (customer.meterType === MeterType.INACTIVE) {
+      balanceInfo = [{ number: UNDEFINED_AMOUNT, label: 'Remaining Balance', unit: '' }];
+      return (
+        <OutlinedCardList info={balanceInfo} primary={false} readOnly />
```
wangannie: You should turn the `primary` prop to be optional defaulting to true because you're setting it to false more often than true.

Also, it's rather unintuitive that the prop that turns the text red is labelled `primary` since in most other contexts, primary represents orange. I'd rename this to something more representative of it's actual behavior (maybe `highlighted` or something like that?)

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -63,105 +62,180 @@ function CustomerProfile(props: CustomerProps) {
...
         }}
-        className={classes.buttonSecondary}
-        startIcon={<ListAltOutlinedIcon />}
-        disableElevation={true}
+        size="small"
+        className={classes.buttonPrimary}
       >
-        Records
-      </Button>
+        <AddIcon style={{ color: 'white' }} />
```
wangannie: You don't need inline styles here because your `color: 'white',` in the `buttonPrimary` class already takes care of that.

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -1,30 +1,26 @@
...
 import BaseScrollView from '../../components/BaseComponents/BaseScrollView';
 import OutlinedCardList, { CardPropsInfo } from '../../components/OutlinedCardList';
 import { CustomerRecord, MeterReadingRecord, PaymentRecord, SiteRecord } from '../../lib/airtable/interface';
 import { selectCurrentCustomer, selectMeterReadingsByCustomerId, selectPaymentsByCustomerId } from '../../lib/redux/customerData';
-import { EMPTY_CUSTOMER } from '../../lib/redux/customerDataSlice';
+import { EMPTY_CUSTOMER, MeterType } from '../../lib/redux/customerDataSlice';
 import { RootState } from '../../lib/redux/store';
 import { selectCurrentSiteInformation } from '../../lib/redux/siteData';
 import { getAmountBilled, getCurrentReading, getPeriodUsage, getStartingReading, getTariffPlanByCustomer } from '../../lib/utils/customerUtils';
+import Button from '../../components/Button';
```
wangannie, line 17: Remove the unused `currentSite` on line 55

src/screens/Customers/CustomerProfile.tsx
```diff
@@ -36,17 +32,20 @@ const styles = (theme: Theme) =>
...
-    buttonSecondary: {
-      borderRadius: '12px',
-      color: 'white',
-      backgroundColor: theme.palette.primary.main,
+    meterInfoGrid: {
+      display: 'flex',
+    },
+    meterInfoCol: {
+      width: '50%',
+      padding: '5px',
```
wangannie, line 43: This padding is also causing this grid to be slightly inset. I couldn't come up with an immediate solution but I've consulted  @mylesdomingo
image

wangannie (approved):
THIS IS GREAT. So much cleaner ✨  Thank you for being patient with this one.

I noted some _final_ cleanup/renaming things but I think it's ready to go once those are addressed. I'll let you know if I get a response from Myles on the padding bug.

PR #69, opened 2021-04-03
[Inventory] Integrate camera functionality for purchase request receipts

wangannie:
[//]: # "These comments are meant for your reference. They are invisible and don't need to be deleted!"

# What's new in this PR
This PR integrates the camera functionality established in #53 with the purchase request creation workflow. Users can take pictures and upload receipts along with their purchase requests.

Note: I ended up separating out CameraButton into a dedicated component since it's repeated several times throughout the app.

## How to review
On the backend, run the latest version of `annie/purchase-request-camera` to test.

**Help fix Annie's bug**
UPDATE: fixed with something in this https://stackoverflow.com/questions/50575215/access-react-router-history-stack
I haven't been able to figure out a smooth navigation workflow with the camera and camera/preview screens. After submitting the purchase request, I used to have it just do `history.goBack()` but now that goes back to one of the camera screens. I've tried:
- a series of `replace`'s (instead of the `push` actions happening in the camera flow currently) but that also makes things messy within the actual camera flow (e.g. if everything is replace instead of push, if you go back on the preview screen, you're kicked all the way back out into the inventory item screen)
- using `history.go(-3)` to go back 3 layers (to bypass the camera screens) but that doesn't work if the user goes through the camera flow multiple times (i.e. to re-take a picture)

**Make a purchase request**
- Go through the whole purchase request creation flow and upload a picture!
  - Make sure the image is uploaded in Airtable

### Online sources
https://reactrouter.com/web/api/history

### Related PRs
Integrates work from #53 
Companion backend PR: https://github.com/calblueprint/meepanyar-node/pull/21

[//]: # "Optional - related PRs you're waiting on/ PRs that will conflict, etc; if this is a refactor, feel free to add PRs that previously modified this code"

## Next steps
- Update styling on the purchase request page
- Look into storage implications of keeping the photo as a URI on the frontend, since it could be a potential bottleneck down the road (see [comment](https://github.com/calblueprint/meepanyar/pull/69#pullrequestreview-627571118))

### Screenshots
image
localhost_3000_inventory(Galaxy S5)

CC: @julianrkung

[//]: # 'This tags Julian as a default. Feel free to change, or add on anyone who you should be in on the conversation.'

PR #73, opened 2021-04-05
Fix blank text field bug

wangannie:

# What's new in this PR
A quick one-line fix for a bug introduced in #69 where default value on textfields was set to `''`making them un-editable for uncontrolled TextFields that don't rely on the optional `value` prop. Now, all `TextFields` should have normal behavior.

Note: this is a temporary fix. We still have this error with controlled/uncontrolled components (for both TextField and Checkbox) that we will handle in a future PR.
image

## How to review
- Make sure you can type into the Inventory Update text field
- Make sure you can use all the text fields in Edit Customer

### Online sources
https://material-ui.com/components/text-fields/#uncontrolled-vs-controlled

### Related PRs
Fixing a bug introduced in #69 

## Next steps
- Convert all text fields to controlled components.

### Screenshots
n/a

CC: @julianrkung

### DIFF
--- a/src/components/TextField.tsx
+++ b/src/components/TextField.tsx
@@ -36,7 +36,7 @@ function Field(props: TextFieldProps) {
         InputProps={{ style: { fontSize: 14 }, disableUnderline: true, autoComplete: 'off' }}
         InputLabelProps={{ shrink: true }}
         color={props.primary ? 'primary' : undefined}
-        value={props.value ? props.value : ''}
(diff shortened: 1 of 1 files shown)


julianrkung (approved):
Would you mind putting that this is to work around a warning? Code looks kinda mysterious without explanation.

Other than that LGTM, and thanks for the proactive fix! Sorry for letting this slide through, I remembered @cnnmon mentioning something about this during the edit customer PR, but I think I checked it working for addCustomer rather than editCustomer.

PR #75, opened 2021-04-06
Customer Payment Functionality and Remove AddToOfflineCustomer

julianrkung:

# What's new in this PR
This PR implements the customer payment functionality and does a rough draft of the frontend payments screen. The frontend is subject to change, and the focus of the PR was to get functionality working for payments.

This PR also changes `editCustomer` to be `updateCustomer` to be consistent with the API of the `Inventory` feature.

This PR also deletes the `addToOfflineCustomer` function, which was planning to be used to modify requests queued by `workbox-background-sync`. It was deleted because workbox updated the way they stored requests so that they're an `ArrayBuffer` instead of a `BLOB` and therefore can't be decoded anymore. The PR therefore also removes all function calls of `addToOfflineCustomer`.

Changelist:
- Implemented screen to add payments
- Removed `addToOfflineCustomer`
  - Removed logic from functions using `addToOfflineCustomer`
- `createMeterReadingandInvoice` back to generated function because it no longer uses `addToOfflineCustomer` and the endpoint was extraneous anyways
- Renamed all `editCustomer`-related to `updateCustomer` and made the API similar to that of their `Inventory` counterparts.

### Payment Functionality
- Create a new customer via the app
- On the Airtable dashboard, go to the "Customers" table, find your customer, and expand the newly created customer
- Click "Link to a record from "Meter Readings and Invoices" and add a new record with a non-zero "Amount Billed" and "Reading" and set the "Date" to be the current day (you can also optionally fill out "Meter Number" but I don't believe that's checked)
- Log into the app again so that the app pulls from airtable
- Go to your created Customer and see it has an "Amount Owed"
- Record a payment for that customer and make sure the "Amount Owed" lowers properly
- Make sure the payment shows up on Airtable

### Patch regeneration
- Run `npm run generate-schema`
- Run `git status` to see that the generated files were modified
- Run `npm run apply-schema-patches` and make sure the patches are applied cleanly
- Run `git status` and make sure the branch is now clean

### Edit Customer
- Edit a customer (you can create a new customer and edit that or edit an existing one)
- Make sure the edits still persist properly on the frontend and on the backend

### Online sources

### Related PRs

## Next steps
- Make Textfield a controlled component
- Make the "Payment" textfield required 

### Screenshots

CC: @julianrkung


src/screens/Customers/CustomerProfile.tsx
```diff
@@ -94,6 +94,8 @@ function CustomerProfile(props: CustomerProps) {
     return (
       <Button
         variant="contained"
+        component={Link}
```
wangannie: (for line 77) you need to change `minUnits` to reflect your schema change! and also update schema patch

wangannie: Also for Remaining Balance (and possibly all floats) you probably want to round it! 
image

wangannie: image
Same in customer cards

src/screens/Customers/AddPayment.tsx
```diff
@@ -0,0 +1,99 @@
...
+        <BaseScreen title="Add Payment" leftIcon="backNav">
+            <div className={classes.content}>
+                <div className={classes.amountOwedContainer}>
+                    <Typography variant='body1' gutterBottom>
+                        Remaining Balance
+                    </Typography>
+                    <Typography variant='h1'>
+                        {totalAmountOwed} Ks
+                    </Typography>
+                </div>
```
wangannie: I feel like this can be replaced by the `OutlinedCardList` component with some customization

src/screens/Customers/AddPayment.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+                <form noValidate>
+                    <TextField label={'Amount Paid (Ks)'} id={'new-payment'} primary={true} onChange={handleSetPaymentAmountInput} />
```
wangannie: Makes more sense to make the ID reflect what this actually is, so `amount-paid`

wangannie (approved):
Noted some bugs + quick fixes but otherwise looking good!

src/screens/Customers/AddPayment.tsx
cnnmon: ^ agreed, also the different naming of Remaining Balance vs Amount Owed is a little confusing (though that's kind of nitpicky)

cnnmon (approved):

src/screens/Customers/AddPayment.tsx
julianrkung: Now using `OutlinedCardList` for this. `remainingBalance` is a better name, thanks!

src/screens/Customers/CustomerProfile.tsx
julianrkung: I imagine this will be a problem for a lot of places. Let's do the rounding all at once so it's done in a consistent way. This is a really good point though, I opened an issue for it, thanks!

PR #76, opened 2021-04-06
[Translation Support] Script to create resources object from Airtable translations DB

micahtyong:
# In this PR 
This PR formalizes [task 1](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#0edcc4efcc6f4b2298ca27d072994e55) of our [roadmap for translation support](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7) by creating a script that pulls down translations from the Airtable and formats it according to [react-i18next](https://react.i18next.com/guides/quick-start) conventions. This script is run any time a production build is made in order to scrape in translations locally for offline-friendly support. 

## How to review
Upon running `npm run translation-sync`, the output should be the following: 
```
> meepanyar@0.1.0 translation-sync /Users/micahyong/Desktop/blueprint/meepanyar
> node ./src/lib/i18next/getResources.js

Successful pulling in translations
```

Then in `src/lib/i18next/resources.js`, we should see the following output: 
```
export default {
  my: {
    translation: {
      'Cancel': 'ဖျက်သိမ်းမည်',
      'Close': 'ပိတ်မည်',
      'Incident Date (Date of Occurence)': 'ဖြစ်ရပ်သည့်နေ့',
      'Preventative Maintenance': 'ကြိုတင်ထိန်းသိမ်းမှု',
      'Remaining': 'ပမာဏ',
      'Spent [Money]': 'သုံးစွဲငွေ',
      'Submit': 'တင်သွင်းမည်',
      ... <a lot more rows> 
  }
}
```

The results should also be the same as #61 and #57. 

## Relevant Links
- [Design document for translation support](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7) 

### Related PRs
- [Fetching and parsing translations](https://github.com/calblueprint/meepanyar-node/pull/10). This PR can be closed after this one is merged. 
- #57 
- #61 

## Next steps
- Propagating this hook throughout the entire app (essentially comes down to replacing any hardcoded strings with our new `useIntl` hook). 
- Support fill-in-the-blank translations as outlined [here](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#6c221ab8002e4ddaad0c1ee82c21dad8)

CC: @julianrkung

PR #61, opened 2021-03-25
[Translation Support] Automatically detect device language and encoding scheme

micahtyong:
# What's new in this PR

In this PR, I address [step 4](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#222323ac6b0b4ada80b8c801898abf1d) of our [roadmap for translation support](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7) by creating two helper functions that detect a device's language and encoding in `i18next/device/ts`. 

## How to review
Note, the following instructions are for Chrome users only. 
1. Follow the same testing instructions as in #57 on this branch. 
2. [Change the browser's primary language to Burmese](https://support.google.com/chrome/answer/173424?co=GENIE.Platform%3DDesktop&hl=en). Be sure to move `Burmese` to the top of your languages when configuring your Advanced settings. This will make `Burmese` the default language on your browser. 
Screen Shot 2021-03-25 at 2 44 10 PM

3. Reopen your localhost with the Mee Panyar application. The text you applied the custom hook to in step 1 should now be in Burmese Unicode. 

## Relevant Links
- [Design document for translation support](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7) 

### Online sources
- [Detecting Myanmar Zawgyi and Unicode in a Javascript application](https://medium.com/@linmyatko/how-to-detect-myanmar-zawgyi-or-unicode-on-modern-frontend-javascript-apps-28d4a1a4a1eb) 

### Related PRs
- [Fetching and parsing translations](https://github.com/calblueprint/meepanyar-node/pull/10)
- #57 

## Next steps
- Propagating this hook throughout the entire app (essentially comes down to replacing any hardcoded strings with our new `useIntl` hook). 
- Updating the translations Airtable as necessary (currently only supports text in the Adalo app) 
- Changing locale abbreviation to `my` instead of `bur` in [fetch and parse function](https://github.com/calblueprint/meepanyar-node/pull/10)
- Regenerate `resources.js` files whenever doing production builds

### Screenshots
After following the steps above and restarting Chrome, the results remain the same as #57. 
Screen Shot 2021-03-25 at 2 43 05 PM

CC: @julianrkung

PR #79, opened 2021-04-08
[Translation Support] Modify hook to support variable input

micahtyong:
# In this PR 
This PR addresses [the pass-by-value challenge](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#f545439216a24965bba433d722fd6648) of our [roadmap for translation support](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7) by adding an optional argument, `input`, which will be used to handle variable inputs.

The important thing to note about this PR is that it relies [these DB constraints](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#e2bec25623f24f29b3fa26bb30049bd7) in the translation DB. 

## How to review
1. Choose any frontend screen or component with some variable input. In this example, we will use `CustomerCard.tsx`. 
2. Import and initialize the translation hook (same as #57) with `import { useInternationalization } from '../lib/i18next/translator';` at the top of the file and `const intl = useInternationalization();` in the component. 
3. Let's localize the text, `Last Updated: {props.date}` be replacing that line with the following (line 107 atow).
```
<Typography className={classes.updatedText}>{intl("Last Updated [Date]", props.date)}</Typography>
```
4. If you want to test this in Burmese, change your browser language to Burmese or modify `device.ts#detectLanguage` to return "my".
5. Run `npm start` and you should see the changes reflected on the customer cards `http://localhost:3000/customers`. Here's one example: 

Screen Shot 2021-04-07 at 8 42 56 PM

**Note**: Some customer cards will not have enough vertical padding to display the full Burmese text. This is out of scope for this PR, but I thought I would point it out. 

## Relevant Links

### Related PRs
- #57 
- #61 
- #76  

## Next steps
- Propagating this hook throughout the entire app (essentially comes down to replacing any hardcoded strings with our new `useIntl` hook). 

CC: @julianrkung


julianrkung (approved):
Followed the steps and it worked, great work! :)

Question: In the event we want `No Readings` translated also, we would have to have translated before passing it as the input param? So in the example you gave in the description, `props.date` would need to have already been translated some time before (or we could alternatively do some call like `intl("Last Updated [Date]", intl(props.date))` ) right?

image

micahtyong:
Yup, you're totally right! I've pushed up another commit that translates the input within the hook itself, so no additional changes need to be done on the implementation side of things. Here's what the output might look like now: 
Screen Shot 2021-04-08 at 9 56 32 AM

PR #80, opened 2021-04-08
Re-style base components

wangannie:

# What's new in this PR
Updated styling and added some error handling for:
- Button: resting, outlined, text variants with disabled states
- TextField: resting, filled, error, disabled states + support for units, required, and input types (e.g. limit to numeric input)
- CameraButton: standardized styling
- SearchBar (formerly `UserSearchBar`): cleaned up, updated styling

Updated color styles to match Figma:
- Secondary is now grey
- Light orange is now `primary.light`
- Removed `info` since it overlapped with the `error` color

I created a BaseComponentsDemo page to easily preview the different states. We'll keep this for now as a reference but delete before handoff.

Note: new component updates are not implemented yet and will be included in a later PR.

## How to review
- Navigate to http://localhost:3000/base-components and see that components match the designs

Help:
image
I can't figure out why this X button on the search bar is BLUE

### Online sources
- Docs for all components in https://material-ui.com/

### Related PRs
- Needs to be merged before #70 

## Next steps
- Integrate changes from components from this PR (e.g. add `required`, `type='number'`, placeholder text, and more to existing instances)
- Cleanup the login page which doesn't actually use our base components
- Clean up all custom hex codes and text styles to match the material ui theme styles

### Screenshots

CC: @julianrkung


src/components/CameraButton.tsx
```diff
@@ -34,19 +33,14 @@ function CameraButton(props: CameraButtonProps) {
...
           <img style={{ width: '100%' }} src={photoUri} />
         ) : (
           <div>
-            <Typography color="primary">
-              <PhotoLibraryIcon />
-            </Typography>
-            <Typography variant="h2" color="primary">
-              Add Photo
-            </Typography>
+            <Typography variant="button" color="primary">+ Add Photo</Typography>
```
julianrkung: There's now an error state for photos. Don't know if you wanted to add this into the PR as well or not

image

julianrkung (approved):
Looked good to me! We'll need to do a big frontend adjustment to get all the screens correctly styled after these base components but I can take charge on that if you don't want to.

I really appreciate you doing this, it'll help a lot when we begin finalizing all the frontend for all the screens :) thank you so much

src/components/CameraButton.tsx
wangannie: Updated to add this!

PR #83, opened 2021-04-09
Site Data Refactor to make similar to CustomerData and InventoryData

julianrkung:

# What's new in this PR
Customer Data and Inventory Data have a nice redux structure that allows for selector use and entity adapters. This PR refactors siteData so that it has the same structure. This was done now because we will begin implement functionality around objects that live in `SiteData` such as Financial Summaries and Tariff Plans.

### Changelog
1. Made `siteData` structure identical to `inventoryData` and `customerData`
2. `siteData` now has 2 entity adapters per site: 1 for Financial Summaries and 1 for tariff plans
3. Removed all instances of `currentSite` and switched them to use the selector `selectCurrentSite`

## How to review
1. Make sure changing sites still work
2. Make sure all pages that used to use `currentSite` are still reachable
  - Home, Customer Profile, Financial Summary (though not used right now), Confirm Payment
2. Pointers on any of the refactored code would be nice as well

### Online sources

### Related PRs

## Next steps
1. Selectors for SiteData to get Tariff Plans based on meter type

### Screenshots
Evidence of no more `currentSite`
image

CC: @julianrkung


src/lib/redux/siteData.ts
```diff
@@ -1,12 +1,22 @@
...
-const getCurrentSiteId = (): string => {
-  return store.getState().siteData.currentSite.id;
+export const getCurrentSiteId = (): string => {
+  return store.getState().siteData.currentSiteId;
 }
 
-export { setCurrentSite, getCurrentSiteId };
+export const selectCurrentSiteId = (state: RootState) => state.siteData.currentSiteId;
+
+export const selectCurrentSite = createSelector(
```
wangannie: Rename to `selectCurrentSiteInformation`?

src/lib/redux/siteData.ts
```diff
@@ -1,12 +1,22 @@
...
+  selectCurrentSiteId,
+  store.getState,
+  (siteId, state) => state.siteData.sites[siteId].siteInformation)
+
+// Returns all SiteRecord[] information
+export const selectAllSites = (state: RootState) : SiteRecord[] => Object.values(state.siteData.sites).map(site => site.siteInformation);
```
wangannie: Also rename to `selectAllSitesInformation`?

Also, why did you access `state.siteData.sites` directly? It seems like our pattern so far has been to use the entity adapter's `getSelectors` and build custom selectors off of those, so I'm curious why you did this instead

src/lib/utils/customerUtils.ts
```diff
@@ -36,9 +37,8 @@ export const getAmountBilled = (currReading: MeterReadingRecord) => {
 };
 
 // TODO: add better error handling
-export const getTariffPlan = (customer: CustomerRecord, currentSite: SiteRecord): TariffPlanRecord | undefined => {
-  const tariffPlanId = customer.tariffPlanId;
-  return currentSite.tariffPlans.find((plan: TariffPlanRecord) => plan.id === tariffPlanId);
+export const getTariffPlan = (customer: CustomerRecord): TariffPlanRecord | undefined => {
```
wangannie: Rename to something like `getTariffPlanByCustomer`? You could also just turn this into a selector similar to `selectProductByInventoryId`.

src/lib/redux/siteData.ts
```diff
@@ -1,12 +1,22 @@
...
 };
 
-const getCurrentSiteId = (): string => {
-  return store.getState().siteData.currentSite.id;
+export const getCurrentSiteId = (): string => {
+  return store.getState().siteData.currentSiteId;
 }
 
-export { setCurrentSite, getCurrentSiteId };
+export const selectCurrentSiteId = (state: RootState) => state.siteData.currentSiteId;
```
wangannie: Is there a reason we're keeping both of these?

src/lib/redux/siteDataSlice.ts
```diff
@@ -1,19 +1,47 @@
...
+// Returns customers in the context of the current site
+export const {
+  selectEntities: selectAllTariffPlans,
+  selectAll: selectAllTariffPlansArray,
+  selectById: selectTariffPlanById,
+  selectIds: selectTariffPlanIds
+} = tariffPlanAdapter.getSelectors(
+  (state: RootState) => state.siteData.sites[state.siteData.currentSiteId].tariffPlans);
+
+// Returns customers in the context of the current site
```
wangannie: Financial summaries **

src/lib/redux/siteDataSlice.ts
```diff
@@ -1,19 +1,47 @@
...
-import { createSlice } from '@reduxjs/toolkit';
-import { SiteRecord, CustomerRecord, FinancialSummaryRecord } from '../airtable/interface';
+import { createEntityAdapter, createSlice, EntityState } from '@reduxjs/toolkit';
+import { RootState } from './store';
+import { SiteRecord, FinancialSummaryRecord, TariffPlanRecord, SiteId } from '../airtable/interface';
+
+const tariffPlanAdapter = createEntityAdapter<TariffPlanRecord>();
+const financialSummaryAdapter = createEntityAdapter<FinancialSummaryRecord>();
+
+// Returns customers in the context of the current site
```
wangannie: Tariff Plans**

wangannie (commented):
I'm a little confused about the new structure with full TariffPlan objects under individual sites since AFAIK a tariff plan can still be shared across multiple sites right? Wouldn't this lead to conflicts with editing tariff plans? I thought we landed on the original plan with sites only tracking tariff plan ids and referencing a shared list of tariff plans that we keep as the single source of truth.

Besides that, I left some nitpicks + renaming suggestions. Everything else looks good!

src/lib/redux/siteData.ts
julianrkung: Will rename to that. I originally had it at that name, but thought I might be breaking convention since others are like `selectAllCustomers` etc

I didn't know a better way to do it. `state.siteData.sites.siteInformation` (only tariffPlans and financialSummaries) isn't an adapter, so there aren't pre-given selectors for it.

src/lib/redux/siteData.ts
julianrkung: For some reason I initially decided against changing all the existing `getCurrentSiteId` to `selectCurrentSiteId(store.getState())`. I don't think there was a good reason for my choice, so I just changed all the `getCurrentSiteId` instances to use the selector and deleted `getCurrentSiteId`

src/lib/utils/customerUtils.ts
julianrkung: That's a better name, thanks. Will do `getTariffPlanByCustomer` first. Might be good to change it to a selector once the `CustomerProfile` PR is finished (don't wanna cause more conflicts than necessary)

julianrkung:
>I'm a little confused about the new structure with full TariffPlan objects under individual sites since AFAIK a tariff plan can still be shared across multiple sites right? Wouldn't this lead to conflicts with editing tariff plans? I thought we landed on the original plan with sites only tracking tariff plan ids and referencing a shared list of tariff plans that we keep as the single source of truth.

Discussed on slack. Confirmed with NPO that tariff plans are bound to only a single site. We both agreed that adapters are a good solution with this constraint

wangannie (approved):
LGTM! Thanks for addressing all of my notes.

PR #84, opened 2021-04-10
[Translation Support] Sync labels between Adalo and Mee Panyar application for translation

micahtyong:
# In this PR 
This PR [syncs labels between the Adalo and Mee Panyar application](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#663a9ea60a5e48cbbb1751fe84400c8b) outlined in our [roadmap for translation support](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7) by adding a script to `getResources.js` which formats a TypeScript object with available words for translations, storing the result in `i18next/words.ts`. 

## How to review
1. Run `npm run translation-sync` to pull in the latest translations and words. 
2. The new React hook is in `src/lib/i18next/translator.ts`. Import it into a file of your choosing (see example below). 
3. The new word bank is in `src/lib/i18next/words.ts`. Import it as a variable called `words`. 
2. Initialize the hook with `const intl = useInternational();`. 
3. Rather than using a hard-coded string, use text from the word bank with `{intl(words.<insert word key here>)}`. Then upon rendering, you should see the results displayed in your browser's language. 

## Relevant Links

### Related PRs
- #57 
- #61 
- #76  
- #79 

### Screenshots

For example, in `CustomerMain.tsx`, we can import our hook and object with
```
import { useInternationalization } from '../../lib/i18next/translator'; 
import words from '../../lib/i18next/words';
```
then use it in `CustomerMain` with 
```
const intl = useIntl(); 
...
<h1 className={classes.title}>{intl(words.customers)}</h1> // there should be some autocomplete for this!
...
```

If the user's default language setting is Burmese and their encoding is in Unicode, they will get the following result: 
Screen Shot 2021-04-10 at 3 54 37 PM

## Next steps
- Propagating this hook throughout the entire app (essentially comes down to replacing any hardcoded strings with our new `useIntl` hook and `words` object). 

CC: @julianrkung


src/lib/i18next/getResources.js
```diff
@@ -43,18 +43,42 @@ const getAllRecords = (table, filterByFormula = '', sort = []) => {
...
+ * Format TypeScript object with available words for translations, storing in ./words.ts. 
+ */
+const getAndSetWordMap = async () => {
+  try {  
+    const resources = require('./resources'); 
+    const words = {}; 
+    for (const key in resources["my"]["translation"]) {
+        const word_key = key.toLowerCase().replace(/ /g,"_").replace(/[^a-z_A-Z]+/g, '');
+        words[word_key] = key; 
+    }
```
julianrkung, line 70: NIT: Can you put a small comment explaining what the code does? I understand it because I have context, but I imagine it'd be hard to read the regex / understand what's going on without context.

src/lib/i18next/words.ts
```diff
@@ -0,0 +1,261 @@
...
+  "mpys_share_profit": "MPY's share [Profit]",
+  "outstanding_payments": "Outstanding [Payments]",
+  "payment_rate_": "Payment Rate (%)",
+  "submitted_automatically_on_date": "Submitted automatically on [date]",
+  "to_approve_inventory_purchase": "To Approve [Inventory Purchase]",
+  "approved_inventory_purchases": "Approved [Inventory Purchases]",
+  "billed_money": "Billed [Money]",
+  "collected_money": "Collected [Money]",
+  "filing_report": "Filing report...",
+  "the_deadline____of_the_current_period__has_passed_a_report_will_be_filed_and_the_site_will_move_to_the_next_reporting_period": "The deadline ([ ])  of the current period ([ ])has passed. A report will be filed and the site will move to the next reporting period."
```
julianrkung, line 260: Small question: This is something that's expect to be fixed in the Adalo DB right rather than with with the translation script right?

julianrkung (commented):
I just realized this could potentially break production builds if a change is made to the Translation DB and isn't fixed in the codebase (since the english key might change). 

We should probably just not rerun the translation sync script automatically during production builds and instead make them manually rerun the translation script and commit the new resources.js and words.ts that get generated so they have a chance to fix whatever errors pop up. What do you think? We should address the ^ problem with this PR.

src/lib/i18next/words.ts
micahtyong: Yeah, your understanding is right--words.ts is updated based on what's in the Adalo DB which we treat as our source of truth.

julianrkung (approved):
LGTM! Thanks :) VERY nice way to get code completion and will definitely save future devs from headaches, good idea on this one!

PR #85, opened 2021-04-11
Tiffany/customer overview

cnnmon:

# What's new in this PR

Updated customer overview & listing to match the Figma, both with styling & functionality. Added a tab component for filtering customers by all/meter/payment/done and updated search feature.

## How to review

- Log in as `testUser`
- Navigate to [http://localhost:3000/customers](http://localhost:3000/customers)
- Click on tabs & search any keywords!

## Next steps

- The search bar doesn't properly clear previous searches, e.g. if you type stuff, close, and re-open the search bar, it won't be empty. This doesn't really mess with functionality, just not ideal.
- Total owed numbers still need to be rounded.
- Did we want to keep a sorting feature / search by meter number?
- Couldn't figure out how to get icons next to labels in tabs :(

### Screenshots

image

image

image

image

CC: @julianrkung


src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,119 @@
...
-  }, [sortBy, filterBy, searchValue]);
+  const allCustomers = useSelector(selectAllCustomersArray) || [];
+  const toMeterCustomers = useSelector(selectCustomersToMeter) || [];
+  const toCollectCustomers = useSelector(selectCustomersToCollect) || [];
+  const customersDone = useSelector(selectCustomersDone) || [];
 
-  const getCustomers = () => {
-    let allCustomers = fullCustomers;
+  const defaultCustomers: CustomerMenu = {
+    all: allCustomers || [],
```
julianrkung: NIT: no need to do the `|| []` twice since you already did it on lines 57-60

src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -51,11 +66,14 @@ export interface HeaderProps {
   name?: string;
   email?: string;
   backAction?: () => void;
+  searchAction?: any;
+  searchExit?: any;
```
julianrkung: Don't use `any` at all if you can avoid (in this case you can). This should be something like:

```
  searchAction?: (e: React.ChangeEvent<HTMLInputElement>) => void;
  searchExit?: () => void;
```

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,119 @@
...
-        <FormHelperText>{filterLabels[1]}</FormHelperText>
-        {filteredCustomersAlt.map((customer, index) => (
-          <Link key={index} to={`${props.match.url}/customer`} onClick={() => setCurrentCustomerIdInRedux(customer.id)} >
-            <CustomerCard name={customer.name} amount={customer.outstandingBalance} date={getLatestReadingDate(customer)} active={customer.isactive} />
-          </Link>
-        ))
-        }
-      </BaseScrollView>
+    <BaseScreen rightIcon="user" title="Customers" searchAction={handleSearchChange} searchExit={exitSearch}>
+      <CustomersTabMenu customers={customers} url={props.match.url} />
```
julianrkung: Wrap everything under a `BaseScrollView` like it had before to prevent weird scrolling behavior shown [here](https://www.loom.com/share/1f21ab1e72054580aba73ec662c65b0a)

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,119 @@
...
+  }
+  const [customers, setCustomers] = useState<CustomerMenu>(defaultCustomers);
 
-    if (searchValue !== '') {
-      allCustomers = allCustomersTrie.get(searchValue);
-    }
-    allCustomers.sort(sortByFunction);
-    sortAndFilterCustomers(allCustomers);
+  //initial setup of trie
+  useEffect(() => {
```
julianrkung: Could you clean this code up (I know you didn't write it, but would be good since we're refactoring)? It looks like we don't need the state `allCustomersTrie` since it'll only ever contain all customers. We also don't need this `useEffect`. We could just create an `allCustomersTrie` in the main function, fill it with all the customers, and always just use that (no state needed, no useEffect needed)

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,119 @@
...
+    if (searchValue !== '') {
+      const currentCustomers = allCustomersTrie.get(searchValue);
+      const newCustomers: CustomerMenu = {
+        all: currentCustomers,
+        toMeter: customers.toMeter.filter((customer: CustomerRecord) => currentCustomers.includes(customer)),
+        toCollect: customers.toCollect.filter((customer: CustomerRecord) => currentCustomers.includes(customer)),
+        done: customers.done.filter((customer: CustomerRecord) => currentCustomers.includes(customer)),
+      }
+      setCustomers(newCustomers);
+    } else {
```
julianrkung: Search is kinda unresponsive / laggy. Is there a way we can make it a bit faster? Maybe you could see if creating a set of `id`s that pass the filter and doing `.has` on the set using the customer ids instead of `.includes` on the customer array would help? 

O(n^2) --> O(n)!!!!! Pog!

Also maybe doing the filtering on just the active tab open might help? Unsure what else would speed it up though

src/screens/Customers/components/CustomersTabMenu.tsx
```diff
@@ -0,0 +1,118 @@
...
+        <Typography variant="body1">Status: Active</Typography>
+      </div>
+      <BaseScrollView>
+        <TabPanel className={classes.tabPanel} value="0">
+          {props.customers.all.map((customer: CustomerRecord, index) => (
+            <Link key={index} to={`${props.url}/customer`} onClick={() => setCurrentCustomerIdInRedux(customer.id)}>
+              <CustomerCard name={customer.name} meterNumber={customer.meterNumber} amount={customer.outstandingBalance} active={customer.isactive}
+              status={getCustomerStatus(customer)}/>
+            </Link>
+          ))}
```
julianrkung: Logic should be abstracted to a helper function since it's repeated for all the different customer types

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,122 @@
...
+}
+
+function CustomerCard(props: CardProps) {
+  const { classes } = props;
+
+  const getStatusIcon = () => {
+    if (props.status === CustomerStatus.METER) {
+      return ( <FlashOnIcon className={classes.icon} /> );
+    } else if (props.status === CustomerStatus.PAYMENT) {
+      return ( <AttachMoneyIcon className={classes.icon} /> );
```
julianrkung: Looks like `Icon` sizes are supposed to be changed via `fontSize` and doing `fontSize='inherit'`, not by setting `height` and `width`

Source: https://material-ui.com/api/icon/

src/screens/Customers/components/CustomersTabMenu.tsx
```diff
@@ -0,0 +1,118 @@
...
+    return undefined;
+  }
+
+  return (
+    <TabContext value={value}>
+      <Tabs className={classes.tabs} textColor="primary" indicatorColor="primary" value={value} onChange={changeTab}>
+        <Tab className={classes.tab} label="All" value="0" />
+        <Tab className={classes.tab} label="Meter" value="1" />
+        <Tab className={classes.tab} label="Payment" value="2" />
+        <Tab className={classes.tab} label="Done" value="3" />
```
julianrkung: You can create add the icon to the label via like this:
(not the best way to name / style. Make an icon class, make a better name, etc... but here's the general gist)


image


image

src/screens/Customers/components/CustomersTabMenu.tsx
```diff
@@ -0,0 +1,118 @@
...
+      </Tabs>
+      <div className={classes.activeContainer}>
+        <div className={classes.active}></div>
+        <Typography variant="body1">Status: Active</Typography>
```
julianrkung: NIT: body1 is default no need to specify defaults

https://material-ui.com/api/typography/

src/screens/Customers/components/CustomersTabMenu.tsx
julianrkung: If the logic could be abstracted to a helper function, it might be short enough to just put in `CustomerMain`. Not sure if you want to do that, but it'd be an option and you wouldn't need to pass `url`. Your call though, I think both solutions are fine. Whatever you think is cleanest! :)

julianrkung (commented):
Gave some feedback! Great job on the whole, this is a real relief HAHAHA. Thanks for all the work you're putting in I really appreciate it!!

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,121 @@
...
+}
+
+function CustomerCard(props: CardProps) {
+  const { classes } = props;
+
+  const getStatusIcon = () => {
+    if (props.status === CustomerStatus.METER) {
+      return ( <FlashOnIcon className={classes.icon} fontSize='inherit' /> );
+    } else if (props.status === CustomerStatus.PAYMENT) {
+      return ( <AttachMoneyIcon className={classes.icon} fontSize='inherit' /> );
```
julianrkung: The flash icon should take you directly to the meter reading page for the customer. The money icon should take you directly to making a payment for the customer.

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,216 @@
...
+        </Tabs>
+        <div className={classes.activeContainer}>
+          <div className={classes.active}></div>
+          <Typography>Status: Active</Typography>
+        </div>
+        <BaseScrollView>
+          {getTabContent(customers.all, "0")}
+          {getTabContent(customers.toMeter, "1")}
+          {getTabContent(customers.toCollect, "2")}
+          {getTabContent(customers.done, "3")}
```
julianrkung: Actually, I think it's better to just do a single

`getTabContent()` here, and just selectively show which customers to show in that function. 

I have the following:

```
  const getTabContent = () => {
      let shownCustomers;
      if (value === '0') {
        shownCustomers = customers.all
      } else if (value === '1') {
        shownCustomers = customers.toMeter
      } else if (value === '2') {
        shownCustomers = customers.toCollect
      } else {
        shownCustomers = customers.done
      }

      return <div>
        {shownCustomers.map((customer: CustomerRecord) => (
          <Link key={customer.id} to={`${match.url}/customer`} onClick={() => setCurrentCustomerIdInRedux(customer.id)}>
            <CustomerCard name={customer.name} meterNumber={customer.meterNumber} amount={customer.outstandingBalance} active={customer.isactive}
            status={getCustomerStatus(customer)}/>
          </Link>
        ))}
      </div>
    };
```

Inside of `BaseScrollView` I have 

```
        <BaseScrollView>
          {getTabContent()}
        </BaseScrollView>
```

I think this is cleaner, and I don't actually notice any noticeable speed differences (though we were already pretty slow to begin with).

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  return (
+    <Card className={classes.singleCard} variant="outlined">
+      <div className={classes.cardContent}>
+        <div className={classes.indicator}>
+          {props.active ? <div className={classes.active} /> : <div className={classes.notActive} />}
+        </div>
+        <div>
+          <Typography variant="h2" className={classes.titleText}>{props.meterNumber}, {props.name}</Typography>
+          <Typography variant="body1" className={classes.totalText}>Total owed <span className={classes.numberText}>{props.amount} Ks</span></Typography>
```
julianrkung: NIT: You shouldn't need `variant='body1'` since it's defautl value

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,121 @@
...
+      display: 'flex',
+      alignItems: 'center',
+      paddingLeft: '17px',
+    },
+    totalText: {
+      color: theme.palette.text.secondary,
+    },
+    titleText: {
+      fontSize: '19px',
+    },
```
julianrkung: Don't use specify fontsize. typography `h2` should already adjust font size for you appropriately

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { Card, CardActions, Typography, IconButton } from '@material-ui/core';
+import ArrowForwardIosIcon from '@material-ui/icons/ArrowForwardIos';
+import AttachMoneyIcon from '@material-ui/icons/AttachMoney';
+import FlashOnIcon from '@material-ui/icons/FlashOn';
+
+const styles = (theme: Theme) =>
+  createStyles({
+    arrow: {
+      color: theme.palette.text.secondary,
+      fontSize: '15px',
```
julianrkung: Can we just remove this and for the `ArrowForwardIosIcon` put `size='small'`?

https://material-ui.com/api/icon/

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,121 @@
+import React from 'react';
+import { createStyles, Theme, withStyles } from '@material-ui/core/styles';
+import { Card, CardActions, Typography, IconButton } from '@material-ui/core';
+import ArrowForwardIosIcon from '@material-ui/icons/ArrowForwardIos';
+import AttachMoneyIcon from '@material-ui/icons/AttachMoney';
+import FlashOnIcon from '@material-ui/icons/FlashOn';
+
+const styles = (theme: Theme) =>
+  createStyles({
+    arrow: {
```
julianrkung: I actually don't see a big difference if i delete this class for `ArrowForwardIosIcon` and instead just put `size='small'`. I think it's the correct color by default, but lmk if I'm wrong.

julianrkung (commented):
Added some comments! Overall looks good, I think we can have Annie review for CSS-related things after this iteration, I can add her to reviewer after giving it another once over once you make the changes

Other notes:
Bottom of Customer Main cuts off cards
image

Long names create issues after the second line. We should ask Ashley how these should be dealt with (I think she'll probably want to make the font smaller)

image

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,121 @@
...
+  classes: {
+    arrow: string; cardContent: string; singleCard: string; totalText: string; numberText: string; divSpacing: string; active: string; notActive: string; arrowSpacing: string; indicator: string; icon: string; titleText: string;
+  };
+}
+
+export enum CustomerStatus {
+  METER = 'Meter' as any,
+  PAYMENT = 'Payment' as any,
+  DONE = 'Done' as any,
+}
```
julianrkung: It would be better to put this somewhere more accessible. Maybe `customerData.ts` instead because it's not intuitive that this enum should live here

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,139 @@
...
+        </IconButton>
+      );
+    } else if (props.status === CustomerStatus.PAYMENT) {
+      return (
+        <IconButton
+          component={Link}
+          to={{
+            pathname: `${props.match.url}/customer/edit`,
+          }}
+          size="small"
```
julianrkung: These should go to`/customers/customer/meter-readings/create` and `/customers/customer/payments/create` respectively. 

These also need to go to the correct customer by setting the `setCurrentCustomerIdInRedux` to this customer like how the `Link` in `Card` is set up in this component

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,139 @@
...
+      height: '7px',
+      width: '7px',
+      borderRadius: '50%',
+      backgroundColor: theme.palette.secondary.light,
+    },
+    notActive: {
+      height: '7px',
+      width: '7px',
+      borderRadius: '50%',
+      backgroundColor: theme.palette.text.disabled,
```
julianrkung: Can we combine this and have `backgroundColor` based on the `props.active` passed in?

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,139 @@
...
+      color: theme.palette.error.main,
+      marginLeft: '5px',
+    },
+    divSpacing: {
+      paddingLeft: '15px',
+    },
+    singleCard: {
+      width: '100%',
+      height: '93px',
+      border: '1px solid ' + theme.palette.text.disabled,
```
julianrkung: NIT: string template (use backtick and `${theme.palette.text.disabled}` instead of adding string

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,139 @@
...
+    icon: {
+      borderRadius: '8px',
+      color: 'white',
+      padding: '5px',
+      backgroundColor: theme.palette.primary.main,
+      fontSize: '30px',
+    },
+  });
+
+interface CardProps {
```
julianrkung: `CustomerCardProps`

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,139 @@
...
+  name: string;
+  customerId: string;
+  meterNumber: number;
+  amount: string;
+  active: boolean;
```
julianrkung: I think it might make more sense to just pass the entire customer here so the props are cleaner? Kinda indifferent though, your call!

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  return (
+      <Card className={classes.singleCard} variant="outlined">
+        <Link to={`${props.match.url}/customer`} onClick={() => setCurrentCustomerIdInRedux(props.customerId)} className={classes.cardContent}>
+          <div className={classes.indicator}>
+            {props.active ? <div className={classes.active} /> : <div className={classes.notActive} />}
+          </div>
+          <div>
+            <Typography variant="h2">{props.meterNumber}, {props.name}</Typography>
```
julianrkung: This should actually be the `customerNumber` not `meterNumber`

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,231 @@
...
     if (searchValue !== '') {
-      allCustomers = allCustomersTrie.get(searchValue);
+      const searchedCustomersArray = allCustomersTrie.get(searchValue);
+      const searchedCustomers = new Set<string>(searchedCustomersArray.map((customer: CustomerRecord) => customer.id));
+
+      const newCustomers: CustomerMenu = {
+        all: searchedCustomersArray,
+        toMeter: customers.toMeter.filter((customer: CustomerRecord) => searchedCustomers.has(customer.id)),
+        toCollect: customers.toCollect.filter((customer: CustomerRecord) => searchedCustomers.has(customer.id)),
+        done: customers.done.filter((customer: CustomerRecord) => searchedCustomers.has(customer.id)),
```
julianrkung: Sorry I just remembered this, but the NPO wanted the search to be on both the `customerNumber` and `name` (so if you input a number it searches by customer number, if input a name it searches by name).

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,231 @@
...
-          </Link>
-        ))
-        }
-      </BaseScrollView>
+    <BaseScreen rightIcon="user" title="Customers" searchAction={handleSearchChange} searchExit={exitSearch}>
+      <TabContext value={value}>
+        <Tabs textColor="primary" indicatorColor="primary" value={value} onChange={changeTab}>
+          <Tab className={classes.tab} label="All" value="0" />
+          <Tab className={classes.tab} label={getMeterTabLabel()} value="1" />
+          <Tab className={classes.tab} label={getPaymentTabLabel()} value="2" />
```
julianrkung: It probably makes more sense to have `value` be based off of the `CustomerStatus` enum rather than `0, 1, 2` so it's easy to understand what tab is what value at a glance.

julianrkung (commented):
Realized I missed some things on my first go around. After these changes are implemented, it looks pretty good to me, so once it's ready for review again can you add @wangannie as a reviewer for her to take a look at anything styling-related and we can merge? I think it's realistic to get this in before bug bash since the changes are pretty minor at this point, great work :)

src/screens/Customers/components/CustomerCard.tsx
cnnmon: should I do some error handling or should I assume that everyone has a customerNumber?

julianrkung: Something basic would be good just so the app doesn't error, but this should be a required field and all records should have it. I jsut changed the Airtable records so all Customers have a unique customerNumber

src/screens/Customers/CustomerMain.tsx
```diff
@@ -128,7 +130,15 @@ function CustomerMain(props: CustomerMainProps) {
 
   const getCustomers = () => {
     if (searchValue !== '') {
-      const searchedCustomersArray = allCustomersTrie.get(searchValue);
+      let searchedCustomersArray;
```
julianrkung, line 112: The more I think about it, the more I think we can just use `useMemo` somewhere here to speed up the slow tabbing / calculation processes. I'm not sure where though and we have more important to fill out, so it might be wise to just come back to put a TODO explaining that this could probably be sped up using React's useMemo, provide a link, and come back later.

src/components/BaseComponents/BaseScreen.tsx
```diff
@@ -8,7 +8,7 @@ const styles = () =>
       top: '85px',
       bottom: '56px',
       width: '100%',
-      height: 'auto',
+      height: '70%',
```
julianrkung: I think changing this line from `auto` makes some of the UI on other screens look weird. The scroll is a problem on other components as well and is an issue Annie has to deal with. She probably has some advice (I think we might be able to just add padding / margin to the bottom of the list so that it doesn't clip?

Link to conversation about whitespace: https://github.com/calblueprint/meepanyar/pull/87#issuecomment-820041876

cc @wangannie is there an easy fix you'd recommend for the last CustomerCard clipping?

Screenshot of funky UI bc of this change:
image

julianrkung (commented):
LGTM but I want @wangannie  to take a look at the styling since I'm not confident on evaluating it. I didn't run into any issues when reviewing, but the search / tab is noticeably slow which I think is something we might have to just live with (I've searched and couldn't find an adequate solution. A lot of people have complained to MUI).

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,243 @@
...
+      let shownCustomers;
+      if (tabValue === CustomerStatus.METER) {
+        shownCustomers = customers.toMeter;
+      } else if (tabValue === CustomerStatus.PAYMENT) {
+        shownCustomers = customers.toCollect;
+      } else if (tabValue === CustomerStatus.DONE) {
+        shownCustomers = customers.done;
+      } else {
+        shownCustomers = customers.all;
+      }
```
wangannie: Nit: this would be cleaner with a Switch. See example in [inventory](https://github.com/calblueprint/meepanyar/pull/87/files#diff-d978fdb45ac563cea4c479598d2e526e48580cb993a6dfbd8566de7610533f94R53-R63)

src/lib/redux/customerData.ts
```diff
@@ -15,6 +15,13 @@ import { selectCurrentSiteId } from './siteData';
 import { RootState, store } from './store';
 import { createSelector } from '@reduxjs/toolkit';
 
+export enum CustomerStatus {
+  ALL = 'All' as any,
+  METER = 'Meter' as any,
+  PAYMENT = 'Payment' as any,
+  DONE = 'Done' as any,
+}
+
```
wangannie: Is there a reason you need `as any`? It looks like our other enums don't use this

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,243 @@
...
-        <FormHelperText>{filterLabels[1]}</FormHelperText>
-        {filteredCustomersAlt.map((customer, index) => (
-          <Link key={index} to={`${props.match.url}/customer`} onClick={() => setCurrentCustomerIdInRedux(customer.id)} >
-            <CustomerCard name={customer.name} amount={customer.outstandingBalance} date={getLatestReadingDate(customer)} active={customer.isactive} />
-          </Link>
-        ))
-        }
-      </BaseScrollView>
+    <BaseScreen rightIcon="user" title="Customers" searchAction={handleSearchChange} searchExit={exitSearch}>
+      <TabContext value={tabValue.toString()}>
```
wangannie: Why exactly is `TabContext` necessary? It seems to work when I comment it out, and the value is already going through the `Tabs` component.

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,243 @@
...
+    searchBar: {
+      zIndex: 1,
+      position: 'absolute',
+      marginTop: '-20px',
+      width: '90%',
+      height: '20px',
+      backgroundColor: 'white',
+    },
+    tab: {
+      color: `${theme.palette.divider}`,
```
wangannie: Remove this -- the color is `textSecondary` by default, which actually matches the Figma.

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,243 @@
...
-
-enum SortBy {
-  NAME = 'Name (A - Z)' as any,
-  METER = 'Meter Number' as any
-}
+    searchIcon: {
+      color: theme.palette.text.secondary,
+      marginRight: '40px',
+    },
+    searchBar: {
```
wangannie: `searchBar`, `searchIcon`, `headerWrapper`, and `title` all aren't being used. Make sure to clean up your classes!

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,243 @@
...
+                />
+              </div>
+            ))}
+        </div>
+      );
+    };
+
+  const getMeterTabLabel = () => (
+    <div>
+      <FlashOnIcon fontSize='inherit' className={classes.tabIcon} />
```
wangannie: Don't need `fontSize='inherit'` since you're overriding it with your class anyways

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,243 @@
...
+          onChange={changeTab}
+          variant="scrollable"
+        >
+          <Tab className={classes.tab} label="All" value={CustomerStatus.ALL} />
+          <Tab className={classes.tab} label={getMeterTabLabel()} value={CustomerStatus.METER} />
+          <Tab className={classes.tab} label={getPaymentTabLabel()} value={CustomerStatus.PAYMENT} />
+          <Tab className={classes.tab} label="Done" value={CustomerStatus.DONE} />
+        </Tabs>
+        <div className={classes.activeContainer}>
+          <div className={classes.active}></div>
```
wangannie: This colored div thing is super jank haha could you replace this with a circle icon? `FiberManualRecord` might work.

src/screens/Customers/CustomerMain.tsx
```diff
@@ -1,228 +1,243 @@
...
+  const classes = styles(props);
+  const { match } = props;
   const [searchValue, setSearchValue] = useState<string>("");
-  const fullCustomers: CustomerRecord[] = useSelector(selectAllCustomersArray) || [];
+  const [tabValue, setTabValue] = useState<CustomerStatus>(CustomerStatus.ALL);
+  const changeTab = (event: React.ChangeEvent<{}>, newValue: CustomerStatus) => {
+    setTabValue(newValue);
+  };
+
+  //default customer arrays
```
wangannie: Nit: could you add spaces after the `//` and capitalize just to make it look cleaner?

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+      <Card className={classes.singleCard} variant="outlined">
+        <Link to={`${props.match.url}/customer`} onClick={() => setCurrentCustomerIdInRedux(customer.id)} className={classes.cardContent}>
+          <div className={classes.indicatorContainer}>
+            <div className={classes.indicator} />
+          </div>
+          <div>
+            <Typography variant="h2">{customer.customerNumber}, {customer.name}</Typography>
+            <Typography className={classes.totalText}>Total owed <span className={classes.numberText}>{customer.outstandingBalance} Ks</span></Typography>
+          </div>
+        </Link>
```
wangannie: You wanna wrap all of this in a `CardContent` tag with a `flexDirection: row` which will take care of the padding and flex height for you. You can see an example in the latest inventory cards in #87

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+        {getStatusIcon()}
```
wangannie: This is also an action so you should move it into `CardActions`

src/screens/Customers/components/CustomerCard.tsx
wangannie: There is a lot of overlap with inventory cards, so I'd recommend checking out [this section](https://github.com/calblueprint/meepanyar/pull/87/files#diff-40f74751624d23ba073081b521ec85b80ad0a9a1735d72c67cedad290f10fb59R40-L48) where you can copy a lot of the structure and css.

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <CardActions className={classes.arrowSpacing}>
```
wangannie: `edge="end"` on your icon button should eliminate the need for `arrowSpacing`

src/screens/Customers/components/RecordsTabMenu.tsx
```diff
@@ -8,9 +8,6 @@ import BaseScrollView from '../../../components/BaseComponents/BaseScrollView';
 
 const styles = (theme: Theme) =>
   createStyles({
-    tabs: {
-      background: 'white',
-    },
     tab: {
       color: `${theme.palette.divider}`,
```
wangannie: Remove this -- should be textSecondary which is the default

src/screens/Customers/components/RecordsTabMenu.tsx
```diff
@@ -24,23 +21,23 @@ const styles = (theme: Theme) =>
...
   invoices: MeterReadingRecord[];
   payments: PaymentRecord[];
+  defaultTab?: string;
 }
 
-function TabMenu(props: TabProps) {
+function RecordsTabMenu(props: RecordsTabMenuProps) {
   const { classes, invoices, payments, defaultTab } = props;
-  const [value, setValue] = React.useState(defaultTab);
+  const [value, setValue] = React.useState(defaultTab ? defaultTab : "0");
```
wangannie, line 32: I'm confused why these tab ids are strings if they're just indices anyways -- could you just make them integers?

src/screens/Customers/CustomerRecords.tsx
```diff
@@ -31,7 +31,7 @@ function CustomerRecords(props: CustomerRecordsProps) {
     <BaseScreen leftIcon="backNav">
       <div className={classes.content}>
         <Typography variant="h1">Records</Typography>
-        <TabMenu defaultTab={defaultTab} invoices={invoices} payments={payments} />
+        <RecordsTabMenu invoices={invoices} payments={payments} />
```
wangannie, line 33: Also delete the  unused `defaultTab` variable

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+    singleCard: {
+      width: '100%',
+      height: '93px',
+      border: `1px solid ${theme.palette.text.disabled}`,
+      borderRadius: '8px',
+      textAlign: 'left',
+      display: 'flex',
+      flexDirection: 'row',
+      marginBottom: '10px',
+    },
```
wangannie: If you switch over to using `CardContent`, most of this will be unnecessary

src/styles/ThemeStyles.ts
```diff
@@ -9,6 +9,7 @@ export const theme = createMuiTheme({
     },
     secondary: {
       main: '#757575',
+      light: '#64DD17', /* active green */
```
wangannie: Since `secondary` is grey, it doesn't make sense to make the `light` variant green. Look into other colors that may fit better, like `info` or `success` https://material-ui.com/customization/palette/

src/screens/Customers/components/RecordsTabMenu.tsx
```diff
@@ -24,23 +21,23 @@ const styles = (theme: Theme) =>
...
   const changeTab = (event: React.ChangeEvent<{}>, newValue: string) => {
     setValue(newValue);
   };
 
   return (
     <TabContext value={value}>
-      <Tabs className={classes.tabs} textColor="primary" indicatorColor="primary" value={value} onChange={changeTab}>
+      <Tabs textColor="primary" indicatorColor="primary" value={value} onChange={changeTab}>
```
wangannie, line 39: Could you add a TODO to address the empty state for this screen?

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+  customer: CustomerRecord;
+  match: any;
+  status?: CustomerStatus;
+}
+
+export default function CustomerCard(props: CustomerCardProps): JSX.Element {
+  const classes = useStyles(props);
+  const customer = props.customer;
+
+  const getStatusIcon = () => {
```
wangannie, line 66: Nit: since most of the `IconButton` is the same, could you consolidate the two cases and use inline ternaries?

wangannie: If you try that and it's not any cleaner, feel free to ignore this.

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+        </IconButton>
+      );
+    }
+  }
+
+  return (
+      <Card className={classes.singleCard} variant="outlined">
+        <Link to={`${props.match.url}/customer`} onClick={() => setCurrentCustomerIdInRedux(customer.id)} className={classes.cardContent}>
+          <div className={classes.indicatorContainer}>
+            <div className={classes.indicator} />
```
wangannie: Same as on `CustomerMain` --  would recommend swapping this out with an icon

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const getStatusIcon = () => {
+    if (props.status === CustomerStatus.METER) {
+      return (
+        <IconButton
+          component={Link}
+          to={{ pathname: `${props.match.url}/customer/meter-readings/create` }}
+          size="small"
+          onClick={() => setCurrentCustomerIdInRedux(customer.id)}
+        >
+          <FlashOnIcon className={classes.icon} fontSize='inherit' />
```
wangannie: Again you're overriding the font size with your class so this 'inherit' is necessary

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+}
+
+export default function CustomerCard(props: CustomerCardProps): JSX.Element {
+  const classes = useStyles(props);
+  const customer = props.customer;
+
+  const getStatusIcon = () => {
+    if (props.status === CustomerStatus.METER) {
+      return (
+        <IconButton
```
wangannie: I see that you did this differently than how you did it for the add icons in `CustomerProfile`. I think the way you did it there was cleaner, so I'd recommend trying to replicate that here. It should be more or less the same.

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+const useStyles = makeStyles((theme: Theme) =>
+  createStyles({
+    cardContent: {
+      flexGrow: 1,
+      display: 'flex',
+      alignItems: 'center',
+      paddingLeft: '17px',
+      color: theme.palette.text.primary,
+    },
+    totalText: {
```
wangannie: Don't need a class for text color -- use `color="textSecondary"`

src/screens/Customers/components/CustomerCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+      color: theme.palette.text.secondary,
+    },
+    numberText: {
+      color: theme.palette.error.main,
+      marginLeft: '5px',
+    },
+    divSpacing: {
```
wangannie: Remove unused

src/components/BaseComponents/BaseScreen.tsx
wangannie: Yeah definitely revert back to auto for now. I tried messing with this to find a workaround but I haven't gotten very far. I'll let you know if I come up with something. For now, just stick to adding extra padding or a blank div at the end of the list.

src/lib/redux/customerData.ts
```diff
@@ -143,3 +150,15 @@ export const selectCustomersToCollect = createSelector(
         return customersToCollect;
     }
 )
+
+export const selectCustomersDone = createSelector(
```
wangannie, line 175: I'm not super familiar with the various use cases, but I'm wondering if you could break this selector down into smaller pieces

wangannie (changes requested):
This is really great work! Sorry I went a little overboard with the comments, but I don't want that to make you think this isn't good!! These screens were in desperate need for some major cleanup so thank you for taking that on.

Most of my comments should be quick fixes/cleanup, no major restructuring. I'd be good to approve once the`CustomerCard` is cleaned up!

src/screens/Customers/components/CustomerCard.tsx
cnnmon: I tried to move over stuff but I don't really think CardContent adds much? It might make it even more dense because I have to make new divs & new styling to take care of the left/right sides with an indicator. Idk what you did though

src/screens/Customers/components/RecordsTabMenu.tsx
cnnmon: image

src/screens/Customers/components/CustomerCard.tsx
cnnmon: I don't know how I should be using CardContent to make things work, can you please show me an example?

cnnmon: Idk how to used CardContent with indicator, also I tried taking out some of the values with CardContent and then the righthand side broke

src/screens/Customers/components/RecordsTabMenu.tsx
cnnmon: What do you mean empty state?

src/screens/Customers/components/CustomerCard.tsx
cnnmon: I don't think I can replicate it because the icons are changing

src/lib/redux/customerData.ts
julianrkung: I think this is fine for now. We can put a comment as a future TODO, but I can't think of anything cleaner off the top of my head

src/screens/Customers/components/CustomerCard.tsx
julianrkung: Maybe we could add a TODO to consider this and file a github issue so it doesn't get lost? Just so it's in for bug bash.

julianrkung (approved):
LGTM. We can add TODO to investigate the CardAction portion if it'll be cleaner

PR #71, opened 2021-04-05
[CustomerData Refactor] CustomerData has Payments and MeterReadings as EntityAdapters

julianrkung:
[//]: # "These comments are meant for your reference. They are invisible and don't need to be deleted!"

# What's new in this PR
This PR changes `customerData` so that it's structurally identical to `inventoryData`. This is mostly done so that all slices in the redux store are stored similarly, and so operations done on different slices (such as adding, removing, etc...) look similar logically. The schema was also regenerated because 2 custom objects `meterReadings` and `payments` are sent from the backend.

This refactor was done now because we're about to complete functionality for payments and meter readings

Changelist
- `meterReadings` and `payments` are no longer stored inside of `customer` objects
- `meterReadings` and `payments` have their own EntityAdapters inside of `customerData`
- Schema was regenerated to fix foreign key issues
- Patches were regenerated to include new customer objects `meterReadings` and `payments`
- New selectors were created for `meterReadings` and `payments`
- Files `Home.tsx`, `CustomerMain.tsx`, `CustomerProfile.tsx` and `CustomerUtils.tsx` were refactored so that they used the selectors created for `meterReadings` and `payments` 

[//]: # "Describe what's new in this PR in a few lines. A description and bullet points for specifics will suffice."

## How to review
1. Use the correct backend branch `julian/customer-data-refactor` PR'd [here](https://github.com/calblueprint/meepanyar-node/pull/22)
2. Open the app and clear site data to reset redux
3. Log in and make sure there are no errors on the Home, CustomerMain, CustomerProfile, or Records screens
4. Create a customer and make sure that they are still properly made on the frontend and backend
5. Edit a customer and make sure they're still properly editable
6. Go to the customer's records and make sure that meter readings and payments are populated

[//]: # 'The order in which to review files and what to expect when testing locally'

### Online sources

[//]: # 'Optional - copy links to any tutorial or documentation that was useful to you when working on this PR'

### Related PRs

[//]: # "Optional - related PRs you're waiting on/ PRs that will conflict, etc; if this is a refactor, feel free to add PRs that previously modified this code"

## Next steps
- Refactor sites
- Functionality for payments and meter readings

[//]: # "What's NOT in this PR, doesn't work yet, and/or still needs to be done"

### Screenshots

[//]: # "Add screenshots of expected behavior - GIFs if you're feeling fancy!"

CC: @julianrkung

[//]: # 'This tags Julian as a default. Feel free to change, or add on anyone who you should be in on the conversation.'

PR #86, opened 2021-04-11
Refactor UserData and pull filtered user info with sites

wangannie:

# What's new in this PR
This PR updates UserData to match the redux structure of CustomerData, SiteData, and InventoryData using selectors and entity adapters. With this change, we no longer use `mapStateToProps` anywhere in the app, as we've completely transitioned over to using selectors!

Additionally, this PR pulls basic user info (id, username, name, admin, siteIds) for all users at a given site. The user data is filtered in the backend so only essential details are passed through.

## How to review
Test with `annie/user-data` on the backend.

- Use the Redux inspector to make sure the data looks right.
- Navigate to inventory and select a purchase request to see that the username shows up. If the user doesn't have a username, it will fall back to the user's id
- Feedback on naming conventions

### Online sources

### Related PRs
Backend PR: https://github.com/calblueprint/meepanyar-node/pull/26

Based on the refactoring work from:
SiteData #83 
CustomerData #71 
InventoryData #60 

## Next steps
- Integrate with settings

### Screenshots
n/a

CC: @julianrkung


src/lib/airtable/schema.js
```diff
@@ -39,6 +39,7 @@ export const Columns = {
 	"Sites": {
 		// Custom objects added to the top of sites object to reduce
 		// likelihood of patch conflicts when new columns are added
+		users: {name: `SiteUsers`, type: `custom-object`},
```
julianrkung, line 42: Remember to make patch for this!

src/lib/redux/userDataSlice.ts
```diff
@@ -1,10 +1,22 @@
...
 
+const siteUsersAdapter = createEntityAdapter<UserRecord>();
+
+// Customized selectors for site users
+export const {
+  selectEntities: selectAllSiteUsersArray,
+  selectAll: selectAllSiteUsers,
+  selectById: selectSiteUserById,
+  selectIds: selectSiteUserIds,
+} = siteUsersAdapter.getSelectors((state: RootState) => state.userData.siteUsers);
```
julianrkung: I just realized my use case requires `SiteIds` also (we should only show users that are a part of the current site for this screen). I could probably do this in a later PR also but it'd require your review. If you have the bandwidth, would you mind doing the setup for this? If not, I don't mind doing this if you just want to put a TODO

image

src/lib/airtable/interface.ts
```diff
@@ -23,13 +23,13 @@ export type Row = {
...
-  sites: SiteRecord[];
-  password: string;
-  name: string;
   admin: boolean;
+  email?: string;
+  photo?: Airtable.Attachment[];
+  siteIds?: string[];
+  sites?: SiteRecord[];
+  password?: string;
+  name?: string;
```
julianrkung: Name shouldn't be optional

src/lib/airtable/interface.ts
```diff
@@ -23,13 +23,13 @@ export type Row = {
...
-  siteIds: string[];
-  sites: SiteRecord[];
-  password: string;
-  name: string;
   admin: boolean;
+  email?: string;
+  photo?: Airtable.Attachment[];
+  siteIds?: string[];
+  sites?: SiteRecord[];
+  password?: string;
```
julianrkung, line 32: Can you put a TODO that we should scrub password out? There's no reason to expose it. We'd eventually want to just not deliver to frontend via a `restrictColumn` thing on the write access resolver but right now `login` doesn't use the write access resolver

src/lib/airtable/interface.ts
```diff
@@ -23,13 +23,13 @@ export type Row = {
...
-  photo?: Airtable.Attachment[];
-  siteIds: string[];
-  sites: SiteRecord[];
-  password: string;
-  name: string;
   admin: boolean;
+  email?: string;
+  photo?: Airtable.Attachment[];
+  siteIds?: string[];
+  sites?: SiteRecord[];
```
julianrkung: `sites` should be a `string[]` and we should remove `siteIds`

image


We might have to pass the  User Record we receive from `Login` through schema generator's `fromAirtableFormat` since right now the user record is basically what we'd get from an Airtable call. No need to do this in this PR, but we should be aware via a TODO

julianrkung (approved):
LGTM after removing some properties in the `UserRecord` interface and putting in some `TODOs`. If you're going to do the `siteId`-related change, I can also take another look if you want!

Remember to update the patch for `schema.js` before you merge

src/lib/redux/userDataSlice.ts
```diff
@@ -1,10 +1,22 @@
...
+
+// Customized selectors for users
+export const {
+  selectEntities: selectAllSiteUsersArray,
+  selectAll: selectAllSiteUsers,
+  selectById: selectSiteUserById,
+  selectIds: selectSiteUserIds,
+} = usersAdapter.getSelectors((state: RootState) => state.userData.users);
+
 interface UserFields {
```
julianrkung: Delete `UserFields` and `Users` since they aren't being used anymore

julianrkung (approved):
LGTM! A couple suggestions before merge. Thanks :)

src/lib/redux/userData.ts
```diff
@@ -49,31 +67,35 @@ const checkOnline = (): void => {
...
 
-const getUserId = (): string => {
-  const state = store.getState();
-  let userId = '';
+export const selectCurrentUserId = (state: RootState): string => state.userData.currentUserId;
 
-  if (state.userData && state.userData.user) {
-    userId = state.userData.user.id;
-  }
+export const selectUserIsSignedIn = createSelector(selectCurrentUserId, (currentUserId) => currentUserId !== '');
```
julianrkung: NIT: maybe `selectCurrentUserIsSignedIn`? Since we're using `CurrentUser` for other naming

Issue #74, opened 2021-04-05
Uncontrolled Text Fields and Components


julianrkung:
There are a lot of uncontrolled components in the app (Login textfields, Add Customer text fields, etc...). All components should be made to be controlled components (have their `value` bound to the component's state)

Look at `EditCustomer` as a guide.

Resources:

What are controlled components: [https://stackoverflow.com/questions/42522515/what-are-react-controlled-components-and-uncontrolled-components#:~:text=A Controlled Component is one,props to the controlled component](https://stackoverflow.com/questions/42522515/what-are-react-controlled-components-and-uncontrolled-components#:~:text=A%20Controlled%20Component%20is%20one,props%20to%20the%20controlled%20component). 

Changing uncontrolled to controlled: [https://goshakkk.name/turn-uncontrolled-into-controlled/](https://goshakkk.name/turn-uncontrolled-into-controlled/)

julianrkung:
Closed with https://github.com/calblueprint/meepanyar/pull/123

PR #87, opened 2021-04-11
[Inventory] Update inventory styling and add history feed

wangannie:

# What's new in this PR
- Updated styling for all inventory screens and components
- Updated all inventory-related TextFields to controlled components, working towards fixing #74 
  - This won't appear to work by default -- see notes in How to Review
- Added a history feed to the inventory profile which combines and sorts all inventory updates and purchase requests.
- Added search functionality to inventory and purchase request screens

Minor base component updates and other app-wide changes:
- Added an optional `currency` prop to `TextField` to add the **MMK** to the start of the input field
- Added an optional `staticPreview` prop to the `CameraButton` which disables the button action + styles the label to match the rest of the design.
  - NOTE/TODO: we should consider renaming this base component to make it more intuitive for the static preview use case.
- Modified the `formatDateStringToLocal` to use a shortened ##/##/ format (still location-sensitive though) because the full-length "April 11th, 2021" was very long and consistently wrapping onto new lines.
- Some modifications to `BaseHeader` for search from #88 (i.e. `bigTitle`, `searchPlaceholder` props) to make it more adaptable across use cases.

**TODO before merge**
- [x] Update plain userIds with names/usernames with updates from #86 
- [x] Confirm number on the history feed with designs
- [x] Rename `PurchaseRequests` to either `PurchaseRequestsList` or `PurchaseRequestsMain` (would appreciate feedback on this @julianrkung). I'd make this change after reviews to keep the diffs easier to read.

## How to review
Overall, compare the frontend with the designs and note any major differences. Keep in mind that some differences were approved by Ashley (see threads in the private Slack channel). Some colors aren't exactly matching Figma (e.g. black text on red badges instead of white) because material UI automatically picks the most accessible color combination which I decided to keep.

### Inventory Profile History
1. Open any inventory item to see its profile. 
2. Create an inventory update. When you are returned to the profile, you should see your update on the top of the feed.
3. Create a new purchase request. When you are returned to the profile, you should see your purchase request at the top of the feed with Pending status.

### Admin Access
**As an admin:**
1. In the Airtable base, make sure that `testUser` has the `Admin` field checked.
2. If you navigate to http://localhost:3000/inventory/purchase-requests, you should see the **All** and **Pending** tabs on the top, as well as approve and deny buttons on any pending purchase request cards.
3. Make sure you are able to approve and deny purchase requests directly from the card on the **All Purchases** screen without having to open the detail page.
  - When you test this, verify that the request status is updated both on the UI, in Airtable, Redux, and when you click on the card to open the Purchase Request screen.
4. If you click on a pending purchase request on the list, you should see the approve and deny buttons, and you should be able to review the request as usual.
5. On the main http://localhost:3000/inventory page, if there are pending purchase requests (if there aren't any, make a few), you should see a badge around the **All Purchases** button.

**As a non-admin:**
1. In the Airtable base, un-check the `Admin` field for `testUser`. Reset the redux store and/or clear site data to update the app.
2. In the **All Purchases** screen http://localhost:3000/inventory/purchase-requests, there should not be any tabs on the top, and any pending requests should display a grey hourglass icon instead.
3. Open a pending purchase request -- you should not see approve/deny buttons.

### Form Inputs
On all form screens (CreatePurchaseRequest, CreateInventoryUpdate, AddInventory), verify that you're only able to type numeric inputs (numbers or a decimal point) where appropriate.

**Text Fields are Controlled Components**
1. In `TextField.tsx`, replace line 56 with `value={props.value}`
2. Open the console in the chrome inspector
3. On any form (CreatePurchaseRequest, CreateInventoryUpdate, AddInventory), click on a text field and start typing. You should NOT see the following error.
image
Note: we need to eventually update to `value={props.value}` but we can't do this until we update all TextFields across the app to controlled components.

### Online sources
https://material-ui.com/ component documentation
https://stackoverflow.com/questions/38697630/material-ui-adding-badge-to-a-tab-in-material-ui-tabbar-tabs/46491649

### Related PRs
n/a

## Next steps
- Edit functionality to change a purchase request's status after the fact

### Screenshots

CC: @julianrkung


src/components/TextField.tsx
```diff
@@ -61,6 +62,7 @@ function Field(props: TextFieldProps) {
         error={props.error}
         helperText={props.helperText}
         InputProps={{
+          startAdornment: props.currency ? <InputAdornment position="start">MMK</InputAdornment> : undefined,
```
julianrkung: Do you think the consistency between using Ks vs MMK matters? I see that we use Ks basically everywhere else in the app. I was treating it as the difference between USD and $ or something similar, so probably not, but wanted to throw the idea out.

src/lib/utils/inventoryUtils.ts
```diff
@@ -28,3 +31,33 @@ export const getInventoryLastUpdated = (inventory: InventoryRecord): string => {
...
+};
+
+export const reviewPurchaseRequest = (purchaseRequest: PurchaseRequestRecord, approved: boolean, userId: string) => {
+  const reviewData = {
+    reviewerId: userId,
+    reviewedAt: moment().toISOString(),
+    status: approved ? PurchaseRequestStatus.APPROVED : PurchaseRequestStatus.DENIED,
+  };
+  updatePurchaseRequest(purchaseRequest.id, reviewData);
+  updatePurchaseRequestInRedux({ id: purchaseRequest.id, ...reviewData });
```
julianrkung, line 63: NIT: Can you put a note here that additional authentication against Airtable is done on the backend?

src/screens/Inventory/PurchaseRequest.tsx
```diff
@@ -1,92 +1,114 @@
...
-          {userIsAdmin && purchaseRequest.status == PurchaseRequestStatus.PENDING && (
-            <div>
-              <Button onClick={() => handleSubmit(purchaseRequest, true)} label={'Approve'} />
-              <Button onClick={() => handleSubmit(purchaseRequest, false)} label={'Deny'} />
-            </div>
-          )}
+        <div className={classes.headerContainer}>
+          <InventoryInfo productId={product.id} lastUpdated={getInventoryLastUpdated(purchaseRequest.inventoryId)} />
+          <div className={classes.reviewButtonsContainer}>
+            {userIsAdmin && purchaseRequest.status == PurchaseRequestStatus.PENDING
```
julianrkung: NIT: is there a reason it's double equals here? I was under impression that best practice was triple equals always. We have some instances of double equals in the codebase but it's probably a low priority cleanup

src/screens/Inventory/components/PurchaseRequestCard.tsx
```diff
@@ -11,56 +15,92 @@ import {
...
-      borderRadius: 10,
+      borderRadius: 6,
       display: 'flex',
-      marginBottom: 10,
+      marginBottom: theme.spacing(1),
     },
-    arrowSpacing: {
-      padding: 0,
+    cardActions: {
+      width: 70,
```
julianrkung: More of a question for me so I understand styling better: I realize in this case the pixel width is to align the icons and the text consistently (when there's 1 icon vs 2 icon). When is it generally appropriate to set a pixel width/height?

julianrkung (commented):
Things I noticed while going through the review process:

When going to an Inventory Profile, I saw some whitespace at the bottom of the profile page. This is noticeable on all inventory profiles that require scrolls
image


The date / time is consistently overflowing (on Figma it's just date). I actually don't mind this since I think time might be an important differentiator, but just wanted to point this out.
image

NIT: Scrolling on Inventory Main that shouldn't exist
https://www.loom.com/share/2a4231302eeb47ccbddbb2c3ed91abd7

I believe purchase requests are supposed to be click-able from Inventory Profile (though Inventory Updates do not need to be). They're underlined on Figma to indicate this. I'm impartial on if this is needed functionality and I'm down to cut this as long as we're on the same page.
image

2 things on the purchase request:
1. "Full Name" instead of "testUser" showed on the Purchase Request I created. This seems to happen for any purchase request i create and then view. Purchase Requests made by "Annie" show up fine, just ones created by the user I'm currently logged in as don't show properly. Maybe this is a bug introduced by https://github.com/calblueprint/meepanyar/pull/86 ? **EDIT** No action needed for this. I got confused. `testUser`'s name on Airtable is "Fuill Name", so it is displaying as appropriate
2. I put 0 for the amount spent on this purchase request and it didn't show up as 0, it instead was blank
image

We should standardize floating point at some point. Maybe just arbitrarily at 2 decimal points for everything? Perhaps we could do it [this](https://stackoverflow.com/a/3644302) way?
image

Might've missed some minor code-related stuff.

Overall looks really good :) can probably merge on next iteration

src/screens/Inventory/components/PurchaseRequestCard.tsx
wangannie: Good question. In most cases you shouldn't need to use a fixed px width/height but I had to here so that the right-most column would line up no matter how many icons there are. HONESTLY it was very painful for me to have to resort to this 😢  but I couldn't think of a better way (besides adding a transparent button of the same width ??)
image

wangannie:
A few responses to your [original big comment](https://github.com/calblueprint/meepanyar/pull/87#pullrequestreview-633812881):

--

> When going to an Inventory Profile, I saw some whitespace at the bottom of the profile page. This is noticeable on all inventory profiles that require scrolls

This padding was intentionally added to #80 because some lists were getting cut off. I figured it'd be safer to err on the side of more extra space than risking having things cut off.

--

> NIT: Scrolling on Inventory Main that shouldn't exist

This is an issue on other pages too. I think I know what would fix it but fixing it now would mess up CustomerMain, so I'm not going to do that now. Maybe we can open an issue for it?

-- 

> I put 0 for the amount spent on this purchase request and it didn’t show up as 0, it instead was blank

That’s fixed if you switch value={props.value}  in text field (something we need to do eventually when we fix controlled components)

--
> I believe purchase requests are supposed to be click-able from Inventory Profile (though Inventory Updates do not need to be). They're underlined on Figma to indicate this. I'm impartial on if this is needed functionality and I'm down to cut this as long as we're on the same page.

I went ahead and added that to the latest update.

src/components/TextField.tsx
```diff
@@ -62,8 +62,12 @@ function Field(props: TextFieldProps) {
         error={props.error}
         helperText={props.helperText}
         InputProps={{
-          startAdornment: props.currency ? <InputAdornment position="start">MMK</InputAdornment> : undefined,
-          endAdornment: props.unit? <InputAdornment position="end">{`${props.unit}(s)`}</InputAdornment> : props.error ? <ErrorIcon color="error" /> : undefined,
+          endAdornment:
+          props.unit || props.currency ? (
+            <InputAdornment position="end">{props.currency ? 'Ks' : `${props.unit}(s)`}</InputAdornment>
```
julianrkung, line 67: Your call, but might be cleaner to just have `props.unit` pass in the `(s)` portion since currencies are units and there's no point in the distinction otherwise

julianrkung (commented):
Found a small bug in the UI when approving purchase requests from the "All Purchase Requests" screen and then navigating to the approved request. After that, I'm down to approve so that we can find out anything else wrong during bug bash

Link to video showing bug: https://www.loom.com/share/52627741a62a44719d0fd91d197fc720

julianrkung:
> Rename PurchaseRequests to either PurchaseRequestsList or PurchaseRequestsMain (would appreciate feedback on this @julianrkung). I'd make this change after reviews to keep the diffs easier to read.

I'm a fan of `PurchaseRequestsMain` just to keep it consistent with `CustomerMain` and the like!

julianrkung (approved):
Approving for bug bash! In the case we just merge with the [bug](https://github.com/calblueprint/meepanyar/pull/87#pullrequestreview-636250595), let's file a Github Issue so we can keep track of it!

PR #88, opened 2021-04-12
Tiffany/search

cnnmon:

# What's new in this PR

An update to BaseHeader to include a search bar & functions determining what happens onSearch / onExit.

## How to review

This is a core part of [this PR](https://github.com/calblueprint/meepanyar/pull/85), so it might be easier to run & test the search bar on that.

CC: @julianrkung @wangannie 


src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -48,16 +63,17 @@ export interface HeaderProps {
...
   backAction?: () => void;
+  searchAction?: any;
+  searchExit?: any;
 }
 
 function BaseHeader(props: HeaderProps) {
-  const { leftIcon, title, rightIcon, classes, match, backAction } = props;
-  const currentUser = useSelector(selectCurrentUser);
-  const name = currentUser?.name || '';
-  const email = currentUser?.email || '';
```
wangannie: You'll need to restore these lines

src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -103,27 +118,43 @@ function BaseHeader(props: HeaderProps) {
...
+      </div>
     </div>
   );
 }
 
-export default withStyles(styles)(BaseHeader);
+const mapStateToProps = (state: RootState) => ({
+  name: state.userData.user?.fields.Name || '',
+  email: state.userData.user?.fields.Email || '',
+});
```
wangannie: Remove this and replace with the selectors that got deleted -- this is breaking it
image

wangannie (changes requested):
I didn't get to test the code because of the bug I pointed out (also see Slack). Ping me when it's ready to review again!

src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -36,10 +39,22 @@ const styles = (theme: Theme) =>
...
+    searchBar: {
+      position: 'absolute',
+      width: '100%',
+      padding: '20px',
+      backgroundColor: 'white',
+    },
+    leftTitle: {
+      float: 'left',
+      padding: '0px 25px',
+      fontSize: '30px',
```
wangannie: Instead of setting a custom font size here, condition the variant in the `Typography` instance. Something like this:
`<Typography ... variant={searchAction ? 'h1' : 'h2'}>{title}</Typography>`

I know the effect is the same, but we want to try to stick to the material standardized styles as much as we can.

src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -36,10 +39,22 @@ const styles = (theme: Theme) =>
...
       padding: 0,
     },
+    searchBar: {
+      position: 'absolute',
+      width: '100%',
+      padding: '20px',
+      backgroundColor: 'white',
+    },
+    leftTitle: {
+      float: 'left',
```
wangannie: Not sure what the function of this is, but it has no effect when I comment it out.

src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -36,10 +39,22 @@ const styles = (theme: Theme) =>
...
+      padding: '0px 25px',
+      fontSize: '30px',
+      color: theme.palette.text.primary,
```
wangannie: Don't need this line bc this is the default

src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -103,25 +119,36 @@ function BaseHeader(props: HeaderProps) {
...
   const right = rightIcon ? icons[rightIcon] : null;
 
+  const onSearchExit = () => {
+    setSearchVisibility(false);
+    searchExit();
+  }
+
   return (
-    <div className={classes.root}>
+    <div className={classes.root} style={{ marginTop: props.searchAction ? '20px' : undefined }}>
```
wangannie: Don't introduce inline styles when you already have a class dedicated to this. You can do this conditionally based on props if you switch to `makeStyles` (see how this is done `OutlinedCardList` for an example).

src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -103,25 +119,36 @@ function BaseHeader(props: HeaderProps) {
...
       {header}
       <div className={classes.toolbar}>
         <div className={classes.left}>{left}</div>
-        <div className={classes.right}>{right}</div>
+        <div className={classes.right}>
+          {props.searchAction ? getIcon(() => setSearchVisibility(true), <SearchIcon className={classes.search} />) : null}
```
wangannie: Fun fact: you can use `&&` and `||` in place of ternaries when you only have one case!

So instead of falling back to `null`, you can make this 
```
{props.searchAction && getIcon(() => setSearchVisibility(true), <SearchIcon className={classes.search} />)}
```

src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -103,25 +119,36 @@ function BaseHeader(props: HeaderProps) {
...
+          {right}
+        </div>
         {profileMenu}
       </div>
+      <div className={classes.searchBar} style={{display: searchVisibility ? 'block' : 'none' }}>
+        <SearchBar placeholder="Search for a customer" onSearchChange={searchAction} onSearchExit={onSearchExit} />
+      </div>
```
wangannie: I see that you use a `getSearchBar()` in the version of `BaseHeader` on `tiffany/customer-overview` where I'm testing this. Did you mean to bring that over here too?

src/components/SearchBar.tsx
```diff
@@ -1,21 +1,37 @@
...
 import React from 'react';
+import CloseIcon from '@material-ui/icons/Close';
 
 interface SearchProps {
   onSearchChange: (event: React.ChangeEvent<{ value: unknown }>) => void;
+  onSearchExit?: any;
   placeholder?: string;
 }
 
+const useStyles = makeStyles((theme: Theme) =>
```
wangannie: Remove `(theme: Theme)` since you're not using it

wangannie: Also since this isn't actually depending on the props in any way, I'd convert this to a `withStyles(styles)(SearchBar)` format that we use in most other places that don't depend on props.

src/components/SearchBar.tsx
```diff
@@ -1,21 +1,37 @@
...
-      placeholder={props.placeholder}
-    />
+    <div className={classes.root}>
+      <MaterialTextField
+        name="search"
+        onChange={props.onSearchChange}
+        variant="standard"
+        fullWidth
+        placeholder={props.placeholder}
+        inputRef={input => input && input.focus()}
```
wangannie: This line makes the SearchBar focus, regardless of where it's placed. I see how it's important in the BaseHeader use case, but in other places (this is clear when I ran it on this branch which still has a standalone `SearchBar`), it focuses the search bar upon opening the page, which probably isn't ideal behavior for using this component in other contexts. Maybe make something like `autoFocus` an optional prop that causes this behavior?

src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -49,16 +64,18 @@ export interface HeaderProps {
...
   backAction?: () => void;
+  searchAction?: any;
+  searchExit?: any;
 }
 
 function BaseHeader(props: HeaderProps) {
-  const { leftIcon, title, rightIcon, classes, match, backAction } = props;
+  const { leftIcon, title, rightIcon, classes, match, backAction, searchAction, searchExit } = props;
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
+  const [searchVisibility, setSearchVisibility] = useState(false);
```
wangannie: I think `searchVisible` would be a more intuitive name here

src/components/BaseComponents/BaseHeader.tsx
```diff
@@ -103,25 +119,36 @@ function BaseHeader(props: HeaderProps) {
...
       <div className={classes.toolbar}>
         <div className={classes.left}>{left}</div>
-        <div className={classes.right}>{right}</div>
+        <div className={classes.right}>
+          {props.searchAction ? getIcon(() => setSearchVisibility(true), <SearchIcon className={classes.search} />) : null}
+          {right}
+        </div>
         {profileMenu}
       </div>
+      <div className={classes.searchBar} style={{display: searchVisibility ? 'block' : 'none' }}>
```
wangannie: Again, don't introduce inline styles if you already have a class.

wangannie (commented):
Noted some quick fixes/cleanup things but this is looking really good! Great work.

src/components/BaseComponents/BaseHeader.tsx
cnnmon: is there a way to do this with states?

wangannie: Ahh good point. Not that I know of.

wangannie (approved):
Everything looks good! Thanks for doing this.

PR #89, opened 2021-04-13
My Information and Site Profile Frontend

julianrkung:

# What's new in this PR

This PR builds out the following mostly according to the Figma (linked [here](https://www.figma.com/file/MzpMtoYyVYFn19DpAXHEci/Mee-Panyar?node-id=2969%3A12361)).

Notable deviations from the Figma to keep in mind:
 - Using Buttons instead of checkmarks to submit edits
 - No tabs for the "User Information" screen
 - "User Information Edit" does not have "Active Status" nor "Name".
 - The number of customers per tariff plan is not available information yet, so it is hardcoded as `-` until a lookup is added to airtable
 - "Unused" tariff plans are empty because of ^ as well.
 -  The "My Information" section is not editable, checking with Ashley to see if / how this should be dealt with
 - Non-admins should eventually not be able to make edits / submit here. Checking with Ashley how designs should be for non-admins
 - The top right icon isn't correct. This is actually changed in Tiffany's PR, so I felt it was fine to leave it alone.
 - No Profile pictures for Users or Last Modified

image

This PR builds out the 

## How to review
You'll need to go through flows as both an Admin and Non-admin to check that the backend correctly rejects non-admin requests for certain actions. I recommend you do all the actions first as a non-admin and make sure they all don't change the Airtable, and then switch over to an admin user and do the flow again and make sure stuff changes, and then change them back.

This PR is probably too large to look closely at code (though the code logic isn't difficult), so it's more important to make sure that all the below flows work as intended.

If you run into errors on startup, make sure you're on the correct branches and clear the site data via devtools and try again. Ping in the private channel if problems persist.

Nonadmin credentials (if you don't wanna make an account): 
- Username: j
- Password: j

Admin credentials:
- Username: testUser
- Password: password

1. Use the correct backend branch called `julian/site-profile` located [here](https://github.com/calblueprint/meepanyar-node/pull/27)

### Navigating to the "Profile" screen
1. On the home page, click the "Account" icon on the top right.
2. This should work for both the Admin / Non-Admin 

### Looking at "My Information"
1. Click "My Information"
2. Make sure that the "My Information" screen matches the Figma, except for there should be no "Checkmark" on the top right screen
3. This workflow is the same for Admin / Non-Admin

### Editing Site Information
1. Just for safety, change the Site to the one called "Schwalmtal"
2. Click the Account Icon --> "Site Information" --> "Site Name"
3. Edit the site name to something and press submit
4. Check the Site's name on Airtable. It should be changed if you were an Admin, and should've stayed the same if non-Admin. If the site name changed in Airtable because you were an Admin, change it back for consistency.
5. Navigate to the App's homepage. The site name should be changed regardless (this inconsistency shouldn't be a problem with non-admins since they shouldn't be doing this change in the first place)

### Editing a Tariff Plan
1. Just for safety, change the Site to the one called "Schwalmtal"
2. Click the Account Icon --> "Site Information" --> "Tariff Plan"
3. Click the "No Charge" tariff plan to change (since no Customers are tied to this one)
4. Change the tariff plan however you like
5. Submit the edits
6. Check the Tariff plan on Airtable. It should be changed if you were an Admin, and should've stayed the same if non-Admin.
7. In both cases, the redux store should change the tariff plan to however you edited it. Check that it did so via devtools. I've attached a screenshot to show approximately how to reach the "No Charge" tariff plan
image

### User Information Edits
1. Click the Account Icon --> "Site Information" --> "User Information"
2. Click on any of the user cards
3. Attempt to make one of the users an admin and submit
4. Check if the user you edited is now on Airtable Admin (the column is checked). It should be changed if you were an Admin, and should've stayed the same if non-Admin.

### Online sources

### Related PRs

## Next steps
- Add "Number of Customers" lookup to tariff plans on Airtable and un-hardcode it in settings page
- Confirm with Ashley how non-admins should interact with this workflow
- Add an airtable column for inactive users and create the inactive users tab

### Screenshots

CC: @julianrkung


kyleqhua (approved):
Everything works as expected! Although perhaps we should change the editing colors to a darker/more standard grey? It seems hard to tell that it is something that would be changed since the lighter grey connates a more "disabled" sense? Also, perhaps we shoud block certain screens for non-admins ie they can still reach the screen to make other user admins despite it not doing anything. Just so it would keep some choherency?

micahtyong (approved):
Everything looks good! I agree with what Kyle said. In addition, I wondered what would happen if we changed site name to something random as a non-admin, so I changed it to "Hello" and it threw an error on the console but the frontend remained the same thereby failing silently. I wonder, from a design perspective, a better way to implement this feature? Perhaps a dropdown? 

Other than those design nits, implementation matches the Figma pretty closely. Awesome work!

PR #98, opened 2021-04-15
Tiffany/add customer frontend

cnnmon:

# What's new in this PR

Revised Add Customer page with new fields & styling according to Figma.

## How to review

- Go to http://localhost:3000/customers/create
- Add a customer

## Next steps

EDIT: I'm getting this error when I click on selects:

image

I couldn't really figure out some styling details, like adjusting borderColor & adding placeholders for Select components on Material UI elements - is it okay to leave that alone for now?

Also, you can submit the form with all the required fields empty.

### Screenshots

image

CC: @julianrkung


src/screens/Customers/AddCustomer.tsx
```diff
@@ -4,43 +4,74 @@ import React, { useState } from 'react';
...
+    selectContainer: {
+      padding: '10px 0px',
+    },
+    nameInput: {
+      width: '130%',
+      paddingRight: '10px',
+    },
+    numberInput: {
+      width: '50%',
+    },
```
julianrkung: This is done in `/base-components` (`BaseComponentsDemo.tsx`) more cleanly using flexbox.
image

The associated code for ^ is as follows

image

image


I think we should switch to the flexbox method as it's easier to interpret as well

src/screens/Customers/AddCustomer.tsx
```diff
@@ -78,26 +114,81 @@ function AddCustomer(props: AddCustomerProps) {
...
+                  <MenuItem key={plan.id} value={plan.id}>
+                    <TariffPlanCard tariffPlan={plan} />
+                  </MenuItem>
+                )}
+              </Select>
+            </FormControl>
+          </div>
+          <div className={classes.buttonContainer}>
+            <div className={classes.button}>
+              <Button label={'Add'} onClick={handleSubmit} fullWidth />
```
julianrkung: Could you also pass a `loading` prop that's set to true when someone submits? This is to make sure that those who "Add" multiple times really quickly don't create multiple customers.

src/screens/Customers/AddCustomer.tsx
```diff
@@ -78,26 +114,81 @@ function AddCustomer(props: AddCustomerProps) {
...
+            />
+          </div>
+          <div className={classes.selectContainer}>
+            <FormControl required variant="outlined" className={classes.formControl}>
+              <InputLabel>Tariff Plan</InputLabel>
+              <Select onChange={handleSelectTariffPlan}>
+                {tariffPlans.map((plan) =>
+                  <MenuItem key={plan.id} value={plan.id}>
+                    <TariffPlanCard tariffPlan={plan} />
+                  </MenuItem>
```
julianrkung: This wasn't on the Figma, but the NPO did say they'd like this sorted descending so that the top tariff plan on the select is the one with the most customers, second is the one with the second-most customers, etc

src/screens/Customers/AddCustomer.tsx
```diff
@@ -78,26 +114,81 @@ function AddCustomer(props: AddCustomerProps) {
...
+            </div>
+          </div>
+          <div className={classes.selectContainer}>
+            <FormControl required variant="outlined" className={classes.formControl}>
+              <InputLabel>Meter Type</InputLabel>
+              <Select onChange={handleSelectMeterType}>
+                <MenuItem value={MeterType.ANALOG_METER}>Analog Meter</MenuItem>
+                <MenuItem value={MeterType.SMART_METER}>Smart Meter</MenuItem>
+                <MenuItem value={MeterType.NO_METER}>No Meter</MenuItem>
+              </Select>
```
julianrkung: Adding the label to match what's in InputLabel helped fix the border issue I think you mentioned

source: https://github.com/mui-org/material-ui/issues/14203#issuecomment-764146829
image

Before:
image

After:
image

src/screens/Customers/AddCustomer.tsx
```diff
@@ -78,26 +114,81 @@ function AddCustomer(props: AddCustomerProps) {
...
+            </FormControl>
+          </div>
+          <div className={classes.textContainer}>
+            <TextField
+              label={'Meter Number'}
+              id={'meter-number'}
+              placeholder={'Input'}
+              type="number"
+              onChange={handleMeterNumberInput}
```
julianrkung: I believe we want controlled inputs on all of our `TextField` (read the "Form Input" portion of Annie's description https://github.com/calblueprint/meepanyar/pull/87 ). Can you put `value={...}` for the textFields in this screen?

julianrkung (commented):
Looks good, I think we only need 1 more iteration and we can merge! Asked for a couple quick fixes but this is looking REALLY good :) Thanks for getting this out so quickly!!

src/screens/Customers/AddCustomer.tsx
cnnmon: Isn't it already value={props.value...} in TextField.tsx? I think this works because I can't type non-number stuff into type="number" inputs.

src/screens/Customers/AddCustomer.tsx
```diff
@@ -78,26 +114,81 @@ function AddCustomer(props: AddCustomerProps) {
...
+              disabled={selectedMeterType === MeterType.NO_METER}
+              required={selectedMeterType != MeterType.NO_METER}
```
julianrkung: NIT: `!==` instead of `!=`

src/screens/Customers/AddCustomer.tsx
```diff
@@ -52,16 +77,24 @@ function AddCustomer(props: AddCustomerProps) {
...
   }
 
+  const handleSelectMeterType = (event: React.ChangeEvent<{ value: unknown }>) => {
+    setSelectedMeterType(event.target.value as MeterType);
+  }
+
   // TODO: Add form input validation and error messaging
   const handleSubmit = (event: React.MouseEvent) => {
+    // Set button as loading
+    setLoading(!loading);
```
julianrkung: Set loading to `true` here instead of `!loading` since we'd never want a submit to cancel a load AFAIK

src/screens/Customers/AddCustomer.tsx
```diff
@@ -78,26 +111,102 @@ function AddCustomer(props: AddCustomerProps) {
...
+          </div>
+          <div className={classes.selectContainer}>
+            <FormControl
+              required
+              variant="outlined"
+              className={classes.formControl}
+            >
+              <InputLabel>Meter Type</InputLabel>
+              <Select onChange={handleSelectMeterType} label={'Meter Type'}>
+                <MenuItem hidden value={MeterType.INACTIVE}>Ianactive</MenuItem>
```
julianrkung: typo `Inactive` not `Ianactive`

src/screens/Customers/AddCustomer.tsx
```diff
@@ -78,26 +111,102 @@ function AddCustomer(props: AddCustomerProps) {
...
+            </div>
+          </div>
+          <div className={classes.selectContainer}>
+            <FormControl
+              required
+              variant="outlined"
+              className={classes.formControl}
+            >
+              <InputLabel>Meter Type</InputLabel>
+              <Select onChange={handleSelectMeterType} label={'Meter Type'}>
```
julianrkung: I think we need to bind `value={selectedMeterType} / value={selectedTariffPlanId}` to all the `Select` components on this screen to get rid of this error

image

src/screens/Customers/AddCustomer.tsx
julianrkung: I think you have to do some manual validation handling here to make sure that all fields are filled out. i do something similar (though more trivial) in `EditTariffPlanInformation` if you want a reference. Basically make sure all fields are valid and if so you can set to loading and send the request, otherwise expose an error message to let the user know to fix something.

Right now the Customer is able to submit a completely blank Customer, which shouldn't be allowed to happen

src/screens/Customers/AddCustomer.tsx
julianrkung: I meant that the we should make these controlled comoponents by passing in the appropriate `value` parameter to all instances of TextField here. So For `Meter Number` it should be `value={meterNumber}`, for name it should be `value={customerName}` etc. I think when you do it, it'll first present an error (the one pasted in Annie's PR description), but we'll be fixing that error soon.

src/screens/Profile/components/TariffPlanCard.tsx
```diff
@@ -17,6 +17,7 @@ const styles = makeStyles((theme: Theme) =>
 
 interface TariffPlanProps {
   tariffPlan: TariffPlanRecord;
+  readOnly?: boolean;
```
julianrkung: I don't know if `readOnly` is the best name here. Maybe `editOnClick` might be better? Since in the other workflow when you click it you go to a screen to edit it.

src/screens/Customers/AddCustomer.tsx
```diff
@@ -4,45 +4,70 @@ import React, { useState } from 'react';
 import { useSelector } from 'react-redux';
 import { Redirect, RouteComponentProps, useHistory } from 'react-router-dom';
 import BaseScreen from '../../components/BaseComponents/BaseScreen';
+import BaseScrollView from '../../components/BaseComponents/BaseScrollView';
 import Button from '../../components/Button';
 import Checkbox from '../../components/Checkbox';
 import TextField from '../../components/TextField';
-import { SiteRecord } from '../../lib/airtable/interface';
+import { SiteRecord, TariffPlanRecord } from '../../lib/airtable/interface';
```
julianrkung: Delete any unused imports

julianrkung (commented):
More comments, sorry I missed during initial runthrough. Frontend looks great though, so after some logic-related changes we should be good to merge

julianrkung:
I wouldn't worry about the strictMode error. Apparently it's a thing for MUI https://stackoverflow.com/questions/61220424/material-ui-drawer-finddomnode-is-deprecated-in-strictmode

src/screens/Customers/AddCustomer.tsx
julianrkung: Actually, you can disregard this comment for now (no need to do error handling) since we're talking with NBJC about how possibly using a standardized library for it

src/screens/Customers/AddCustomer.tsx
```diff
@@ -4,45 +4,70 @@ import React, { useState } from 'react';
...
   const { classes } = props;
   const history = useHistory();
 
-  const [selectedTariffPlanId, setSelectedTariffPlanId] = useState("");
   const [customerName, setCustomerName] = useState("");
-  const [hasMeter, setHasMeter] = useState(false);
-  // TODO: @julianrkung Look into constraints on meter number input.
-  const [meterNumber, setMeterNumber] = useState("");
-  const [customerInactive, setCustomerInactive] = useState(false);
+  const [customerNumber, setCustomerNumber] = useState("");
```
julianrkung, line 61: I don't believe the backend API endpoint passes the `customerNumber` field when creating a customer (code [here](https://github.com/calblueprint/meepanyar-node/blob/main/index.js#L72)). Can you make a backend branch so the `customerNumber` assigned here persists to the customer record created (should be a short ~2 line PR)?

julianrkung (approved):
LGTM! Thanks so much ;-;

PR #94, opened 2021-04-15
[Translation Support] Apply translations (screens/Customers)

micahtyong:
# In this PR

In this PR, we follow the steps in the ["Applying Our Hook" section](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#a7e5cbd6962e4dbd850b724c3eee7b18) of our [roadmap for translation support](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#cfc52fc0865e4bb2b140e54d395343ba) in order to translate `screens/Customers` for `meepanyar`. 

## Coverage
- [x] `screens/Customers` e.g. (`CustomerMain.tsx`)
- [ ] `screens/Camera`
- [ ] `screens/FinancialSummary`
- [ ] `screens/Home`
- [ ] `screens/Inventory`
- [ ] `screens/Profile`

## How to Review
1. Change your browser device setting to Burmese (or return "my" in `device.ts#detectLanguage`)
2. Run the application and navigate to the screens corresponding to `screens/Customers`. 
3. Ensure nothing is breaking
4. Test in English and ensure wording is still correct

We will translate the rest of `screens/*` in subsequent PRs. 

cc: @julianrkung

PR #104, opened 2021-04-17
[Translation Support] Apply translations (screens/Inventory)

micahtyong:
# In this PR

In this PR, we follow the steps in the ["Applying Our Hook" section](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#a7e5cbd6962e4dbd850b724c3eee7b18) of our [roadmap for translation support](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#cfc52fc0865e4bb2b140e54d395343ba) in order to translate `screens/Inventory` for `meepanyar`. 

## Coverage
- [ ] `screens/Customers` e.g. (`CustomerMain.tsx`) #94 
- [ ] `screens/Camera`
- [ ] `screens/FinancialSummary`
- [ ] `screens/Home`
- [x] `screens/Inventory`
- [ ] `screens/Profile`

## How to Review
1. Change your browser device setting to Burmese (or return "my" in `device.ts#detectLanguage`)
2. Run the application and navigate to the screens corresponding to `screens/Inventory`. 
3. Ensure nothing is breaking
4. Test in English and ensure wording is still correct

We will translate the rest of `screens/*` in subsequent PRs. 

cc: @julianrkung


src/lib/i18next/device.ts
```diff
@@ -1,29 +0,0 @@
-/**
- * Detect and return user device language.
```
julianrkung: Is this file supposed to be deleted?

tsconfig.json
```diff
@@ -1,26 +0,0 @@
-{
```
julianrkung: Is this tsconfig file supposed to be deleted?

src/screens/Inventory/AddInventory.tsx
```diff
@@ -171,17 +175,17 @@ function AddInventory(props: AddInventoryProps) {
...
+          placeholder={intl(words.eg_x, "5")}
           unit={formik.values.unit}
           type="number"
-          label={'Starting Amount'}
+          label={intl(words.starting_amount)}
           id={'startingAmount'}
           value={formik.values.startingAmount}
           onChange={formik.handleChange}
           error={formik.touched.startingAmount && Boolean(formik.errors.startingAmount)}
           helperText={formik.touched.startingAmount && formik.errors.startingAmount}
```
julianrkung, line 183: I think the `helperText` value `formik.errors.startingAmount` might also need to be translated? @wangannie to make sure.

src/screens/Inventory/AddInventory.tsx
```diff
@@ -148,7 +152,7 @@ function AddInventory(props: AddInventoryProps) {
             <div style={{ marginRight: 8, flex: 2 }}>
               <TextField
                 required
-                label={'New Item Name'}
+                label={intl(words.unit_name)}
                 id={'newProductName'}
                 value={formik.values.newProductName}
```
julianrkung, line 154: This might also need to be run through the translator since it shows the value of `newProductName`

src/screens/Inventory/AddInventory.tsx
```diff
@@ -148,7 +152,7 @@ function AddInventory(props: AddInventoryProps) {
             <div style={{ marginRight: 8, flex: 2 }}>
               <TextField
                 required
-                label={'New Item Name'}
+                label={intl(words.unit_name)}
```
julianrkung: This translates to `Unit Name` not `New Item Name`

src/screens/Inventory/AddInventory.tsx
```diff
@@ -111,9 +114,10 @@ function AddInventory(props: AddInventoryProps) {
   };
 
   const filter = createFilterOptions<string>();
+  // `+ ${intl(words.new_inventory_item)}`
```
julianrkung: nit: delete commented

src/screens/Inventory/CreateInventoryUpdate.tsx
```diff
@@ -61,7 +64,7 @@ function CreateInventoryUpdate(props: CreateInventoryUpdateProps) {
   };
 
   return (
-    <BaseScreen title="Update Item" leftIcon="backNav">
+    <BaseScreen title={intl(words.updated_x, words.item)} leftIcon="backNav">
```
julianrkung: `update_x` probably works better than `updated_x` here

src/screens/Inventory/CreateInventoryUpdate.tsx
```diff
@@ -72,18 +75,18 @@ function CreateInventoryUpdate(props: CreateInventoryUpdateProps) {
...
             required
             type="number"
             unit={product.unit}
-            label={`Updated Amount`}
+            label={intl(words.last_updated_amount)}
             id={'updatedAmount'}
             value={formik.values.updatedAmount}
             onChange={formik.handleChange}
             error={formik.touched.updatedAmount && Boolean(formik.errors.updatedAmount)}
             helperText={formik.touched.updatedAmount && formik.errors.updatedAmount}
```
julianrkung, line 87: Check with annie if we need to translate `formik.errors.updatedAmount`

src/screens/Inventory/CreatePurchaseRequest.tsx
```diff
@@ -117,17 +120,17 @@ function CreatePurchaseRequest(props: CreatePurchaseRequestProps) {
...
             currency
-            label={'Amount Spent'}
+            label={intl(words.amount_spent_paid)}
             id={'amountSpent'}
-            placeholder="e.g. 5"
+            placeholder={intl(words.eg_x, "5")}
             value={formik.values.amountSpent}
             onChange={formik.handleChange}
             error={formik.touched.amountSpent && Boolean(formik.errors.amountSpent)}
             helperText={formik.touched.amountSpent && formik.errors.amountSpent}
```
julianrkung, line 128: Check if `formik.errors...` values should be translated

src/screens/Inventory/CreatePurchaseRequest.tsx
```diff
@@ -117,17 +120,17 @@ function CreatePurchaseRequest(props: CreatePurchaseRequestProps) {
             required
             type="number"
             currency
-            label={'Amount Spent'}
+            label={intl(words.amount_spent_paid)}
```
julianrkung: Does the `words.amount_spent_paid` correctly translate to `Amount Spent` and not `Amount Spent [Paid]` if no input is given to it?

src/screens/Inventory/PurchaseRequest.tsx
```diff
@@ -78,7 +81,7 @@ function PurchaseRequest(props: PurchaseRequestsProps) {
   };
 
   return (
-    <BaseScreen title="Inventory Receipt" leftIcon="backNav">
+    <BaseScreen title={intl(words.inventory_purchase)} leftIcon="backNav">
```
julianrkung: inventory receipt not inventory purchase, though I guess we don't have a translation for this?

src/screens/Inventory/components/InventoryCard.tsx
```diff
@@ -41,9 +44,9 @@ function InventoryCard(props: InventoryCardProps) {
       <CardContent className={classes.cardContent}>
         <div className={classes.leftContent}>
           <Typography  variant="h2">{product.name}</Typography>
-          <Typography variant="body1" color="textSecondary">{`Last Updated: ${lastUpdated}`} </Typography>
+          <Typography variant="body1" color="textSecondary">{`${intl(words.last_updated_amount)}: ${lastUpdated}`} </Typography>
         </div>
-        <Typography variant="h2">{`${currentQuantity} ${product.unit}(s)`}</Typography>
+        <Typography variant="h2">{`${currentQuantity} ${product.unit}(${intl(words.s)})`}</Typography>
```
julianrkung, line 49: Just checking: we're only translating the hardcoded strings, and the dynamic units / names / etc are meant to be in the proper language already right?

src/screens/Inventory/components/InventoryInfo.tsx
```diff
@@ -64,17 +69,17 @@ function InventoryInfo(props: InventoryInfoProps) {
           <Typography variant="h2" color="textPrimary">
             {product.name}
           </Typography>
-          <Typography variant="caption">Last Updated</Typography>
+          <Typography variant="caption">{intl(words.last_updated_date)}</Typography>
           <Typography variant="caption">{lastUpdated}</Typography>
```
julianrkung: Should this be `intl(words.last_updated_date, lastUpdated)` instead?

julianrkung (commented):
Sorry for the late review. Left some comments / questions but didn't get to test the code out. Might leave more comments after trying out the app.

From the looks of it, we don't have the proper translations for a lot of the words we need (ex: Update Item, Inventory Receipt). If we're not able to adequately translate the app right now, it might be better to do a controlled example of how to translate the app and present it to the NPO so that they can do the rest of the translations. What do you think @micahtyong ?

src/screens/Inventory/AddInventory.tsx
micahtyong: What does `newProductName` represent? If this is some variable user input and not a predefined label, then it wouldn't be supported by our translator.

src/screens/Inventory/AddInventory.tsx
micahtyong: Yeah, a brief explanation of what `formik` is doing and the role it plays would be great

src/screens/Inventory/CreatePurchaseRequest.tsx
micahtyong: Good catch, they're different. Thanks!

src/screens/Inventory/components/InventoryCard.tsx
micahtyong: Yup. I translate the letter `s` because it's meant to indicate plurality and it seems to be hardcoded everywhere throughout the app. Are you suggesting I should leave it in English?

src/screens/Inventory/components/InventoryInfo.tsx
micahtyong: Good catch!

src/screens/Inventory/PurchaseRequest.tsx
micahtyong: Just added one. NPO will need to translate it though.

micahtyong:
My goal is to apply the translation hook throughout the app and have the English version remain unchanged; then, it's up to the NPO to go into the metadata DB and properly write in translations. We can use `screens/Inventory` as our model to be safe, if that's what you're thinking (cc: @julianrkung)

src/screens/Inventory/AddInventory.tsx
julianrkung: Ah ok I understand, yes you're right, thanks!

src/screens/Inventory/components/InventoryCard.tsx
julianrkung: No, just checking! This is fine (we can have the NPO address it if desired behavior is different).

src/screens/Inventory/AddInventory.tsx
```diff
@@ -113,7 +116,7 @@ function AddInventory(props: AddInventoryProps) {
   const filter = createFilterOptions<string>();
 
   return (
-    <BaseScreen title="New Inventory" leftIcon="backNav">
+    <BaseScreen title={intl(words.new_inventory_item)} leftIcon="backNav">
```
julianrkung: NIT: According to figma should be `New Inventory` not `New Inventory Item`

src/screens/Inventory/CreateInventoryUpdate.tsx
```diff
@@ -72,18 +75,18 @@ function CreateInventoryUpdate(props: CreateInventoryUpdateProps) {
         </div>
         <form onSubmit={formik.handleSubmit} noValidate>
           <TextField
-            placeholder={'e.g. 5'}
+            placeholder={intl(words.eg_x, "5")}
             required
             type="number"
             unit={product.unit}
-            label={`Updated Amount`}
+            label={intl(words.last_updated_amount)}
```
julianrkung: Should be `Updated Amount`. In this case, `last_updated_amount` is less understandable since this is where the user should be putting the new updated amount

src/screens/Inventory/PurchaseRequest.tsx
```diff
@@ -92,22 +95,22 @@ function PurchaseRequest(props: PurchaseRequestsProps) {
...
           id={'amount-purchased'}
           value={purchaseRequest.amountPurchased}
         />
-        <TextField label={'Amount Spent'} currency disabled id={'amount-spent'} value={purchaseRequest.amountSpent} />
-        <TextField label={'Notes'} disabled id={'notes'} value={purchaseRequest.notes || 'None'} />
+        <TextField label={intl(words.amount_spent_paid)} currency disabled id={'amount-spent'} value={purchaseRequest.amountSpent} />
+        <TextField label={intl(words.notes)} disabled id={'notes'} value={purchaseRequest.notes || intl(words.none)} />
         <TextField
-          label={'Submitted By'}
+          label={intl(words.submitted_by_person)}
```
julianrkung: These show up as `Amount Spent [Paid]` and `Submitted By [Person]` on the screen. Do you think it'd be helpful to have some sort of optional `ignorePlaceholders` parameter that can be passed to `intl` so this can be controlled by the developers more easily? Alternatively we could pass a whitespace string like ` '  ' `  (space is required to pass the truthy value check) but that doesn't seem as clean and would prob require commenting to explain what's going on

image

src/screens/Inventory/components/InventoryCard.tsx
```diff
@@ -41,9 +44,9 @@ function InventoryCard(props: InventoryCardProps) {
       <CardContent className={classes.cardContent}>
         <div className={classes.leftContent}>
           <Typography  variant="h2">{product.name}</Typography>
-          <Typography variant="body1" color="textSecondary">{`Last Updated: ${lastUpdated}`} </Typography>
+          <Typography variant="body1" color="textSecondary">{`${intl(words.last_updated_amount)}: ${lastUpdated}`} </Typography>
```
julianrkung: This should be `last_updated_date` since this typography shows date, not amount

src/screens/Inventory/components/InventoryInfo.tsx
```diff
@@ -40,19 +42,22 @@ const useStyles = makeStyles((theme: Theme) =>
...
 
-const getPurchaseRequestButton = () => (
-  <Link to={'purchase-requests/create'}>
-    <Button label="Purchase" />
-  </Link>
-);
+function PurchaseRequestButton() {
+  const intl = useInternationalization(); 
+  return (<Link to={'purchase-requests/create'}>
+  <Button label={intl(words.inventory_purchase)} />
```
julianrkung: This should just be `Purchase` or else it bloats the button

image

julianrkung (commented):
These are all the changes I saw while reviewing. @wangannie if you could clarify on the `formik` stuff I think Micah would be able to make some changes and we might be able to merge on the next iteration if all the english translations align.

Good work Micah :). The approach you said ^ makes sense and I also think it's the best way forward to complete translations

src/screens/Inventory/AddInventory.tsx
wangannie: Oh yes. Basically the strings in parentheses in `validationSchema` are the error messages/helper text things corresponding to various input requirements, so yeah those would also need to be translated.

src/screens/Inventory/CreateInventoryUpdate.tsx
```diff
@@ -72,18 +75,18 @@ function CreateInventoryUpdate(props: CreateInventoryUpdateProps) {
...
             id={'updatedAmount'}
             value={formik.values.updatedAmount}
             onChange={formik.handleChange}
             error={formik.touched.updatedAmount && Boolean(formik.errors.updatedAmount)}
```
wangannie, line 86: You'll also need to translate the messages in `validationSchema` too

src/screens/Inventory/CreatePurchaseRequest.tsx
wangannie: Yes they should! But in validationSchema. Otherwise it's like this 
image

src/screens/Inventory/components/InventoryInfo.tsx
```diff
@@ -64,17 +69,16 @@ function InventoryInfo(props: InventoryInfoProps) {
           <Typography variant="h2" color="textPrimary">
             {product.name}
           </Typography>
-          <Typography variant="caption">Last Updated</Typography>
-          <Typography variant="caption">{lastUpdated}</Typography>
+          <Typography variant="caption">{intl(words.last_updated_date, lastUpdated)}</Typography>
```
wangannie: I'm not sure what formatting `.last_updated_date` is using but it doesn't seem right in English or Burmese. We intentionally made the date formatting location-sensitive, so that could probably be excluded from the translation.
image

src/lib/i18next/resources.js
```diff
@@ -1,21 +1,74 @@
...
+      "Customers to Meter": "ငွေတောင်းခံရန်ကျန်သောမီးအသုံးပြုသူများ",
+      "e.g. [X]": "e.g. [X]",
       "Edit [x]": "-ကိုပြောင်းပြင်မည်",
+      "Explanation": "Explanation",
+      "Filter by": "Filter by",
+      "Has meter": "Has meter",
       "Incident Date (Date of Occurence)": "ဖြစ်ရပ်ဖြစ်သည့်နေ့",
       "Income and expenses tracked through app": "အက်ပ်မှခြေရာခံသည့်၀င်ငွေနှင့်ကုန်ကျစရိတ်",
+      "Inventory Purchase": "Inventory Purchase",
+      "Inventory receipt": "Inventory receipt",
```
wangannie: Nit: capitalize Receipt

src/lib/i18next/words.ts
```diff
@@ -1,19 +1,72 @@
...
+  "customers_to_meter": "Customers to Meter",
+  "eg_x": "e.g. [X]",
   "edit_x": "Edit [x]",
+  "explanation": "Explanation",
+  "filter_by": "Filter by",
+  "has_meter": "Has meter",
   "incident_date_date_of_occurence": "Incident Date (Date of Occurence)",
   "income_and_expenses_tracked_through_app": "Income and expenses tracked through app",
+  "inventory_purchase": "Inventory Purchase",
+  "inventory_receipt": "Inventory receipt",
```
wangannie: Capitalize R

wangannie (commented):
Julian already noted most of my concerns/questions, and I think the biggest change is just translating all the things in `validationSchema`. Besides that, everything else looked pretty good to me!

src/screens/Inventory/PurchaseRequest.tsx
micahtyong: Based on our agreed-upon [translation rules](https://www.notion.so/calblueprint/Translation-Support-Design-Document-7703018a8c4e403dab8aaa49d2a322e7#e2bec25623f24f29b3fa26bb30049bd7), anything wrapped in brackets should be removed unless the input is explicitly meant to go there. For now, I'll go with the temporary fix you mentioned.

src/screens/Inventory/AddInventory.tsx
```diff
@@ -31,24 +33,12 @@ const styles = () =>
 
 const NEW_PRODUCT_LABEL = '+ New Inventory Item';
```
julianrkung: This also needs to be translated (it shows up at the bottom of the "Select" list and they choose it if they want to create a new type of inventory item.

julianrkung (approved):
Looked over the translated app in burmese and it looks good to me after translating the comment, thanks for getting this done :)

PR #126, opened 2021-04-22
Offline indicators: Snackbar, Modal, Sync icons

wangannie:

# What's new in this PR
This PR adds various indicators to the login, customer, and inventory screens for when the user is offline:
- The `Snackbar` component auto-dismisses after 5 seconds and is used for small reminders/warnings when offline
- The `OfflineDialog` component (modal) is a more interruptive indicator used to block users from taking actions (like editing details, creating purchase requests/inventory updates) on new records that haven't been loaded to Airtable yet. This forces the user to go back instead of viewing the profile screen.
- Small sync icons in cards (`CustomerCard`, `InventoryCard`, `InventoryInfo`, `PurchaseRequestCard`) indicate records that were created offline and have not yet been synced.

**TODO BEFORE MERGE**
- [x] Revert the bottom nav icon (that's currently rigged to take the user offline)

## How to review

For easy testing, I rigged the old **Incidents** icon on the bottom nav to set `isOnline` to `false` in Redux. This only really works for testing a few workflows since it only gives you a few seconds before detecting you're online again, so I'd still recommend testing by going offline in Dev Tools > Application > Service Workers.

image

Overall, keep an eye out for language consistency (i.e. is the snackbar message matching the designs).

**Login Screen**
- Login screen should simply show the snackbar anytime the user is offline.
image

**New Customer**
1. Go offline
2. Create a new customer
3. Find the customer in the list -- it should have a sync icon next to its name, and clicking the card should take you to the profile with a modal that forces you back to the main screen.

**New Inventory**
1. Go offline
2. Create a new inventory item (try both existing items and creating your own product, should both end up the same)
3. Find the new inventory item in the list -- it should have a sync icon next to its name and clicking the card should take you to the profile with a modal that forces you back to the main list screen.

**Reviewing a new purchase request**
1. Go offline
2. For an existing inventory item (not one that was created offline), create a new purchase request.
3. Go to the Purchase Requests screen (http://localhost:3000/inventory/purchase-requests). You should see your new purchase request with a sync icon next to the product name.
4. Try clicking approve/deny on the card (still from PurchaseRequestsMain). It should show the snackbar and the review shouldn't go through.
5. Open the purchase request detail screen by clicking on the card. Try approving/denying again. The behavior should be the same as step 4.
**Note:** you should still be able to create inventory updates and create purchase requests for inventory items _not_ created offline. You should also be able to review purchase requests that weren't created offline.

### Online sources
- https://material-ui.com/api/snackbar/
- https://material-ui.com/components/dialogs/

### Related PRs
n/a

## Next steps
- Add these indicators with other screens where necessary (i.e. Home)
- @micahtyong update the snackbar and dialog messages with translations

### Screenshots

CC: @julianrkung


src/screens/Login.tsx
```diff
@@ -110,6 +114,11 @@ function Login() {
           {/* <Button label="Forgot password?" fullWidth variant="text" /> */}
         </form>
       </div>
+      <Snackbar
+        open={!isOnline}
+        noBottomMargin
+        message="You are not connected to a network. Please reconnect to log in."
+      />
```
julianrkung: What do you think about the snackbar being on indefinitely if you're not online in login? IMO it's confusing otherwise because once the snackbar is gone there's no indication at all that the user is offline (even after a login fails)

src/screens/Inventory/PurchaseRequest.tsx
```diff
@@ -73,15 +76,27 @@ function PurchaseRequest(props: PurchaseRequestsProps) {
...
+    if (!isOnline) {
+      setShowSnackbar(true);
+      // 5 second delay to reset back so it can be shown again.
+      setTimeout(function () {
+        setShowSnackbar(false);
+      }, 5000);
+    } else {
+      reviewPurchaseRequest(purchaseRequest, approved, currentUserId);
+      history.goBack();
+    }
```
julianrkung: I don't think we should use `isOnline` here as the indicator when seeing if the purchase request has an offlineID seems to be the real limiting factor. We can execute the `reviewPurchaseRequest` so long as the `purchaseRequest` doesn't have an offline ID.

With the current workflow, the User isn't able to approve any purchase requests while offline, when I think we really only need to prohibit them from reviewing their own purchase request that they might've made when they were offline

src/screens/Inventory/components/PurchaseRequestCard.tsx
```diff
@@ -62,20 +79,25 @@ export const getPurchaseRequestStatusIcon = (status: PurchaseRequestStatus, size
...
-    purchaseRequest.status = newStatus;
-    reviewPurchaseRequest(purchaseRequest, approved, userId);
+    if (!isOnline) {
+      showSnackbarCallback();
+    } else {
+      const newStatus = approved ? PurchaseRequestStatus.APPROVED : PurchaseRequestStatus.DENIED;
+      setStatus(newStatus);
+      purchaseRequest.status = newStatus;
+      reviewPurchaseRequest(purchaseRequest, approved, userId);
+    }
```
julianrkung: Pasting the comment I made in `PurchaseRequest` since it applies here too

>I don't think we should use isOnline here as the indicator when seeing if the purchase request has an offlineID seems to be the real limiting factor. We can execute the reviewPurchaseRequest so long as the purchaseRequest doesn't have an offline ID.

>With the current workflow, the User isn't able to approve any purchase requests while offline, when I think we really only need to prohibit them from reviewing their own purchase request that they might've made when they were offline

julianrkung (approved):
Looks good! I made a comment on the Login snackbar I wanted to get your feedback on.

The Purchase Request revoew I think should be do-able so long as the purchase request has an Airtable ID. I think the only case where it's NOT doable is if an admin creates their own purchase request and tries to review it (in which case we can throw the "Please reconnect to approve / deny this purchase request")

I'm down to approve since I think the fixes are minor here!

src/screens/Inventory/PurchaseRequest.tsx
wangannie: Switched to only block approvals for requests with offline ids! Good catch.

src/screens/Login.tsx
wangannie: Good call -- that definitely makes more sense in this case.

PR #127, opened 2021-04-22
Preserve '0' input for Purchase Request amount spent

### LINKED ISSUE #115: [BUG] Ability to purchase zero quantity and amount spent in inventory
**Describe the bug**
Users are able to input 0 in "Amount Purchased" and "Amount Spent" when purchasing inventory. However, it produces a strange behavior because the app assumes the 0 is an empty field. For example, after inputting 0 then taking a picture of the receipt, the inputs disappear. If the user still submits with 0's, the end result is an empty field in the Inventory Receipt page.

**To Reproduce**
1. Go to the inventory purchase page
2. Enter 0's for Amount Purchased and Amount Spent
3. Take a picture
4. **Upon return, the 0's entered in the Amounts fields disappears**
5. Enter 0's in the fields again
6. Click Submit
7. See the Recent Updates for the inventory item
8. Click the purchase record
9. **The Amount Purchased and Amount Spent is an empty field**

**Expected behavior**
The input should not accept 0 as a valid input or allow the 0 without the strange empty field behavior.

**Screenshots**
After coming back from taking picture of receipt
image

After submitting with 0's
image

**Device (please complete the following information):**
 - Device Model: Samsung S9
 - OS: Android 10

**Environment:**
Which environment(s) is this bug located in?
local/development
✅ production

CC: @julianrkung

wangannie:

# What's new in this PR

This PR resolves #115 where if users entered '0' in the Create Purchase Request flow, their inputs would disappear after returning from the camera flow. This also changes the minimum "amount purchased" from 0 to restrict to non-zero positive numbers.

The issue was resolved by changing `props.location.state?.amountPurchased as string` to `String(props.location.state?.amountPurchased)`. For some reason `0 as string` which converts it to `''`, whereas `String('0')` keeps it as `'0'`.

## How to review
Test this flow (adapted from the issue -- shoutout to Fred for the helpful level of detail 👏🏻 )
1. Go to the inventory purchase page
2. Enter 0 for Amount Spent
3. Take a picture
4. Upon return, the 0 entered in the Amount Spent field **should not disappear**
5. Click Submit
6. See the Recent Updates for the inventory item
7. Click the purchase record
8. The Amount Spent **should not be** an empty field

### Online sources
n/a

### Related PRs
n/a

## Next steps
n/a

### Screenshots

CC: @julianrkung

### DIFF
--- a/src/screens/Inventory/CreatePurchaseRequest.tsx
+++ b/src/screens/Inventory/CreatePurchaseRequest.tsx
@@ -29,7 +29,10 @@ const styles = (theme: Theme) =>
   });
 
 const validationSchema = yup.object({
-  amountPurchased: yup.number().min(0, 'Please enter a valid amount').required('Please enter an amount'),
+  amountPurchased: yup
+    .number()
+    .positive('Please enter a valid amount greater than 0')
+    .required('Please enter an amount'),
   amountSpent: yup.number().min(0, 'Please enter a valid amount').required('Please enter an amount'),
   notes: yup.string(),
   receipt: yup.string().required('Please upload a receipt image'),
@@ -54,8 +57,8 @@ function CreatePurchaseRequest(props: CreatePurchaseRequestProps) {
 
   const formik = useFormik({
     initialValues: {
-      amountPurchased: (props.location.state?.amountPurchased as string) || '',
-      amountSpent: (props.location.state?.amountSpent as string) || '',
+      amountPurchased: String(props.location.state?.amountPurchased) || '',
+      amountSpent: String(props.location.state?.amountSpent) || '',
       notes: (props.location.state?.notes as string) || '',
       receipt: (props.location.state?.photo as string) || '',
     },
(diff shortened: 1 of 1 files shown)


src/screens/Inventory/CreatePurchaseRequest.tsx
```diff
@@ -54,8 +57,8 @@ function CreatePurchaseRequest(props: CreatePurchaseRequestProps) {
 
   const formik = useFormik({
     initialValues: {
-      amountPurchased: (props.location.state?.amountPurchased as string) || '',
-      amountSpent: (props.location.state?.amountSpent as string) || '',
+      amountPurchased: String(props.location.state?.amountPurchased) || '',
+      amountSpent: String(props.location.state?.amountSpent) || '',
```
julianrkung, line 61: Question:  why was it changed to this? Is there a difference?

julianrkung (approved):
you're SO fast thank you!!

src/screens/Inventory/CreatePurchaseRequest.tsx
wangannie: This is what was causing the issue. For some reason`0 as string` which converts it to `''`, whereas `String('0')` keeps it as `'0'`. I'll note this in the PR description.

PR #129, opened 2021-04-23
[OUTDATED] Fix purchase request reviews from list view

### LINKED ISSUE #117: [BUG] Reject/Approve purchase doesn't go through
**Describe the bug**
When approving/rejecting a purchase of inventory in the All Purchases tab, the state reverts to the no action state when entering the purchase record. The no action state persists when going back to the list, and it looks like no action has been taken.

**To Reproduce**
1. Approve or reject a purchase request
2. Click the purchase request to get more detail
3. **Both the approve and reject buttons are now present (even selecting an action on the item)**
3. Go back to the full list
4. **Both the approve and reject buttons shows up on the request card as well**

**Expected behavior**
When an admin approves or rejects a request from the list, the action should happen. Going into the record's detail page, the record should be approved or rejected. Going back to the full list, the record should be approved or rejected.

**Screenshots**
Initial state
image

Reject the knife purchase
image

After entering into the record detailed page - the rejection is reverted
image

Going back to the full list - both the rejection and approval buttons shows up
image

**Device (please complete the following information):**
 - Device Model: Samsung S9
 - OS: Android 10

**Environment:**
Which environment(s) is this bug located in?
local/development
✅ production

CC: @julianrkung

wangannie:

# What's new in this PR

This PR resolves #117 which was silently blocking purchase request reviews from the list view. It was due to a TypeError caused by attempting to directly modify the `readOnly` `purchaseRequest.status`. My fix involves essentially copying the entire record each time we modify the status.

## How to review
- Feedback on my actual fix would be appreciated -- I'm honestly not super sure what exactly is happening when I'm using `useState` with the record passed in through props, so I'm worried about changes not being saved through the entire workflow, though it has worked with all of my testing so far.

### Online sources
n/a

### Related PRs
n/a

## Next steps
n/a

### Screenshots
n/a

CC: @julianrkung


julianrkung (approved):
Interestingly,  the original error only happens if the User creates the purchase request AND approves it before the Redux store is refreshed from an Airtable pull. I don't know enough about Redux to know why it applies readOnly in that case.

I'm okay with this solution since we're keeping the immutability of purchase request.

The only decent alternative is to have the `/purchase-requests/purchase-request` endpoint grab directly from the redux store rather than getting passed a `purchaseRequest`. I'm fine with this solution also (maybe it's even cleaner? We could keep `status` as the state rather than the entire `purchaseRequest`).

Another uglier option is to delete the `purchaseRequest.status = newStatus` (the line that errors) and pass the updated status state to `Link` if it's clicked, so the state pass will look like

pathname: `/inventory/purchase-requests/purchase-request`, state: { purchaseRequest: {...purchaseRequest, status} }} ... >` which is ugly but the file is simple enough where a comment would clarify the situation.


Your call. I'm okay with this solution since it works and we have more pressing work to do.

wangannie:
Closing since this is replaced by #143 due to changes introduced in #134

PR #131, opened 2021-04-23
[Inventory] Fix offline functionality for new products and inventory

wangannie:

# What's new in this PR

This PR fixes a bug (discovered by @julianrkung) where creating a new product offline (no Airtable id is assigned) made it impossible to create a new inventory and inventory update record corresponding to that product.

The solution is to extract the inventory update creation process out to custom endpoints:
- For new products, this is done in`/products/create`, which now a. creates the product record, b. creates the inventory record, and c. creates the inventory update record.
- For existing products, the `/inventory/create` endpoint now a. creates the inventory record and b. creates the update record.

**Additional changes:**
- Since there was quite a bit of overlap in the work done in `createProductInventoryAndUpdate` and `createInventoryAndUpdate`, I created helper functions `generateInventoryUpdate`, `generateInventory`, and `generateProduct` which do the work of generating a record based on the `EMPTY____RECORD` version.

TODO before merging:
- [x] Update the requests patch

## How to review
Test with `annie/bugfix-offline-new-product` on the backend.

Test all of the following both online and offline:
- In Add Inventory, create a **new product** and make sure that a corresponding inventory record and update are created as well (check Airtable if online, Redux if offline)
- In Add Inventory, select an **existing product** and make sure that a corresponding inventory record and update are created as well (check Airtable if online, Redux if offline)

Feedback requests:
- It feels very weird to have so much overlap between the `createProductInventoryAndUpdate` and `createInventoryAndUpdate`, but I'm not sure how to consolidate them and still preserve the offline behavior.

### Online sources
n/a

### Related PRs
n/a

## Next steps
n/a

### Screenshots
n/a

CC: @julianrkung


src/lib/utils/inventoryUtils.ts
```diff
@@ -61,4 +67,39 @@ export const reviewPurchaseRequest = (purchaseRequest: PurchaseRequestRecord, ap
...
+
+export const generateInventory = (siteId: string, currentQuantity: number, productId: string, id?: string): any => {
+  const inventory = JSON.parse(JSON.stringify(EMPTY_INVENTORY)); // Make a deep copy of an empty record
+  inventory.siteId = siteId;
+  inventory.currentQuantity = currentQuantity;
+  inventory.periodStartQuantity = currentQuantity;
+  inventory.productId = productId;
+  inventory.id = id || '';
+  return inventory;
+};
```
julianrkung: I think these are confusing functions if no comments are added to describe the contract the developer enters when calling this function (ie: if no id is provided upon function call, it is up to the caller to add the ID to this object before storing it into the redux store). 

It's not intuitive at first glance why `id`, which MUST be unique to be able to safely store into the redux store, is an optional parameter here. The purpose of this function might be more clear if you omitted `id` as a parameter altogether and added comments that an id property must be inserted into the object before being able to store it into the redux store.

I'm okay with the optional lD parameter (though I don't love it), but it's necessary to describe what the developer must do after the call to be able to work with this object just like any other inventory object (they have to add a unique ID to it).

src/lib/airtable/request.js
```diff
@@ -209,20 +208,39 @@ export const createManyFinancialSummaries = async (records) => {
...
+// NONGENERATED: Create a product (inventory type), inventory record, and inventory update, then add them to 
+// Redux. Returns the ids of each new record.
+export const createProductInventoryAndUpdate = async (product, startingAmount, siteId, userId) => {
+  let productId = '';
+  let inventoryId = '';
+  let inventoryUpdateId = '';
   try {
     delete product.id; // Remove the id field to add to Airtable
-    productId = await createRecord(Tables.Products, product);
+    const resp = await fetch(`${process.env.REACT_APP_AIRTABLE_ENDPOINT_URL}/products/create`, {
```
julianrkung, line 220: NIT: Can you comment that product, inventory, and inventoryUpdate creation is done on the single endpoint to account for offline functionality?

src/lib/airtable/request.js
```diff
@@ -235,28 +253,35 @@ export const createManyProducts = async (records) => {
...
+// NONGENERATED: Create an inventory record (add product to site) and inventory update, then add both to Redux.
+// Returns the ids for the new records.
+// NOTE: this requires that the site and product already exist in Airtable. If the product doesn't exist,
+// use createProductInventoryAndUpdate.
+export const createInventoryAndUpdate = async (inventory, userId ) => {
   let inventoryId = "";
+  let inventoryUpdateId = "";
+  
   try {
+    delete inventory.id; // Remove the id field to add to Airtable
```
julianrkung, line 269: NIT: Can you comment that inventory, and inventoryUpdate creation is done on the single endpoint to account for offline functionality?

julianrkung (commented):
This flow makes sense! I think it fixes the problem (I had no problems when testing). Can you pull from main so I can test the entire workflow with the new offline snackbars?

Also: I think I can approve after a round of testing with most recent main and either added comments and/or changes to the `generate*` API

Great work by the way! You fixed this bug really quickly!

src/lib/utils/inventoryUtils.ts
wangannie: Removed the optional param and updated comments!

julianrkung (approved):
LGTM! You're so good thank you for fixing !! :^)

PR #135, opened 2021-04-24
Report Listing & Individual Pages

cnnmon:

# What's new in this PR

Built out all reports, report card components, and individual report screens.

Note: I don't know if I should automatically omit the 2021-04 report from All Reports. Is there supposed to be an existing logged financial summary for the current period? I thought this wouldn't exist until the end of the period.

## How to review

http://localhost:3000/reports

ReportsMain -> ReportCard -> Report

## Next steps

Need to add maintenance and incidents data / respective selectors. I just don't think these have been built out yet.

### Screenshots

image

image

CC: @julianrkung


src/components/ListItemWrapper.tsx
```diff
@@ -66,7 +74,7 @@ const ListItemWrapper = (props: ListItemWrapperProps) => {
   const getRightText = () =>
   (<ListItemText
     className={classes.rightTextStyles}
-    primaryTypographyProps={{ align: 'right' }}
+    primaryTypographyProps={{ align: 'right', variant: 'h4' }}
```
julianrkung: This might break other uses of this component. We should pass this as an optional prop like `rightTextVariant` or something 

This makes the settings page right text bolder than on Figma
image

src/components/OutlinedCardList.tsx
```diff
@@ -27,7 +28,7 @@ const useStyles = makeStyles((theme: Theme) =>
     root: (props: CardProps) => ({
       marginTop: '10px',
       border: '1px solid',
-      borderColor: props.readOnly ? theme.palette.background.default : theme.palette.text.disabled,
+      borderColor: props.highlightedBorder ? theme.palette.primary.main : props.readOnly ? theme.palette.background.default : theme.palette.text.disabled,
```
julianrkung: NIT: Can you put parentheses around the second ternary statement so it's easier to read? Otherwise nested ternaries can be pretty confusing to people who aren't used to reading JS

```
props.highlightedBorder ? theme.palette.primary.main : (props.readOnly ? theme.palette.background.default : theme.palette.text.disabled)
```

src/lib/redux/customerData.ts
```diff
@@ -171,6 +171,30 @@ export const selectCustomersToCollect = createSelector(
     }
 )
 
+export const selectTotalAmountBilled = createSelector(
```
julianrkung: the name should specify that this calculates for the current period only

src/lib/redux/customerData.ts
```diff
@@ -171,6 +171,30 @@ export const selectCustomersToCollect = createSelector(
...
+    (customers, state) => {
+        let totalAmountBilled: number = 0;
+        customers.forEach((customer: CustomerRecord) => {
+          totalAmountBilled += selectAmountOwedInCurrentPeriodByCustomerId(state, customer.id) || 0;
+        });
+        return totalAmountBilled;
+    }
+)
+
+export const selectTotalAmountCollected = createSelector(
```
julianrkung: the name should specify that this calculates for the current period only

src/lib/redux/inventoryData.ts
```diff
@@ -136,3 +138,29 @@ export const setCurrentInventoryIdInRedux = (inventoryId: string): void => {
 export const addProductToRedux = (product: ProductRecord): void => {
   store.dispatch(addProduct(product));
 };
+
+export const selectAmountPurchaseRequestedApproved = createSelector(
```
julianrkung: NIT: maybe `selectCurrentPeriodPurchaseRequestsApprovedTotalAmountSpent`?

src/lib/redux/inventoryData.ts
```diff
@@ -136,3 +138,29 @@ export const setCurrentInventoryIdInRedux = (inventoryId: string): void => {
...
+      purchaseRequests.forEach((pr: PurchaseRequestRecord) => {
+        if (pr.status === PurchaseRequestStatus.APPROVED) {
+          amount += pr.amountSpent;
+        }
+      });
+      return amount;
+    }
+)
+
+export const selectAmountPurchaseRequestedDenied = createSelector(
```
julianrkung: NIT: maybe `selectCurrentPeriodPurchaseRequestsDeniedTotalAmountSpent`?

src/lib/redux/siteData.ts
```diff
@@ -34,4 +35,8 @@ export const updateTariffPlanInRedux = (tariffPlan: Partial<TariffPlanRecord>) =
 
 export const updateSiteInRedux = (siteUpdates : Partial<SiteRecord>) => {
   store.dispatch(updateSite(siteUpdates))
-}
\ No newline at end of file
+}
+
+export const selectCurrentSiteProfit = () => {
+  return selectTotalAmountCollected(store.getState()) - selectAmountPurchaseRequestedApproved(store.getState());
```
julianrkung: Selectors can build off other selectors via the `createSelector` function. You can look at `inventoryData` or `customerData` for inspiration, but should look something like

```
export const selectCurrentSiteProfit = createSelector(
selectTotalAmountCollected,
selectAmountPurchaseRequestedApproved,
(totalAmountCollected, totalAmountSpent) => totalAmountCollected - totalAmountSpent
```

src/screens/Reports/ReportsMain.tsx
```diff
@@ -0,0 +1,100 @@
...
+    totalAmountCollected,
+    totalAmountSpent: 0,
+    totalProfit,
+    inventoryAmountApproved,
+    inventoryAmountDenied,
+    period: getCurrentPeriod(),
+    isapproved: false,
+    lastUpdated: '0',
+    issubmitted: false,
+  }
```
julianrkung: To make this cleaner, you could also have this logic be a selector that returns a `FinancialSummaryRecord` and put it in like `siteData`.

src/screens/Reports/ReportsMain.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const reports: FinancialSummaryRecord[] = useSelector(selectAllFinancialSummariesArray) || [];
+
+  const getDeadline = (report: FinancialSummaryRecord) => {
+    const m = moment(new Date(report.period)).endOf('month');
+    const d = moment.duration({'days' : currentSite.gracePeriod});
+    m.add(d);
+    return formatDateStringToLocal(m.toString());
+  }
```
julianrkung: This is a useful function! Can you put it in `momentUtils`? I think I can use this for `Home`

src/lib/redux/siteData.ts
```diff
@@ -34,4 +35,8 @@ export const updateTariffPlanInRedux = (tariffPlan: Partial<TariffPlanRecord>) =
 
 export const updateSiteInRedux = (siteUpdates : Partial<SiteRecord>) => {
   store.dispatch(updateSite(siteUpdates))
-}
\ No newline at end of file
+}
+
+export const selectCurrentSiteProfit = () => {
```
julianrkung: Make sure that this calculates the totalProfit for the period and not total profit over all the periods!

src/lib/redux/inventoryData.ts
```diff
@@ -136,3 +138,29 @@ export const setCurrentInventoryIdInRedux = (inventoryId: string): void => {
...
+  selectAllCurrentSitePurchaseRequestsArray,
```
julianrkung, line 171: Make sure that you calculate the amount denied for just the current period not all periods

src/lib/redux/inventoryData.ts
```diff
@@ -136,3 +138,29 @@ export const setCurrentInventoryIdInRedux = (inventoryId: string): void => {
 export const addProductToRedux = (product: ProductRecord): void => {
   store.dispatch(addProduct(product));
 };
+
+export const selectAmountPurchaseRequestedApproved = createSelector(
+    selectAllCurrentSitePurchaseRequestsArray,
```
julianrkung, line 158: Make sure that you calculate the amount approvedfor just the current period not all periods

julianrkung (commented):
Left comments the frontend is so beautiful ;-; 

Mostly naming NITs, cleanup, and some logic issues!

Really great work I'm so thankful for you ;-;

src/lib/redux/inventoryData.ts
cnnmon: spain

src/lib/redux/siteData.ts
cnnmon: image
tried that but it's giving me this error

src/lib/redux/inventoryData.ts
julianrkung: spain.

src/lib/redux/siteData.ts
julianrkung: okok fair! Could you write a comment saying that `createSelector` isn't being in this case used because the needed selectors aren't initialized yet when the file is read?

src/components/ListItemWrapper.tsx
```diff
@@ -7,6 +7,7 @@ import ArrowForwardIosIcon from '@material-ui/icons/ArrowForwardIos';
 interface ListItemWrapperProps {
   leftText: string;
   rightText?: string;
+  rightTextBlack?: boolean;
```
julianrkung: NIT: `boldRightText`  rather than `black`

src/lib/redux/siteData.ts
```diff
@@ -34,4 +37,48 @@ export const updateTariffPlanInRedux = (tariffPlan: Partial<TariffPlanRecord>) =
...
-}
\ No newline at end of file
+}
+
+// Can't use "createSelector" because selectors aren't initialized before file is read
+export const selectCurrentSiteProfit = (state: RootState) => {
+  return selectCurrentPeriodTotalAmountCollected(state) - selectCurrentPeriodPurchaseRequestsApprovedTotalAmountSpent(state);
+}
+
+export const selectCurrentFinancialSummary = () => {
```
julianrkung: NIT: maybe `selectCurrentPeriodFinancialSummary` might be a better name? Might wanna specify that it's "Current Period"

julianrkung: Also can you make this a selector? You can do that by having it take in a `(state: RootState)` parameter. like the `selectCurrentSiteProfit` above it. That way, you can use the `useSelector` hook [here](https://github.com/calblueprint/meepanyar/pull/135/files#diff-b2c3464f4daed3c11f588a4135b99349b9fea6e991ce234b4174b884b2d35511R32)

src/screens/Reports/ReportsMain.tsx
```diff
@@ -0,0 +1,68 @@
...
+          <ReportCard
+            report={currentReport}
+            deadline={currentDeadline}
+            match={match}
+            current
+          />
+        </div>
+        <div className={classes.section}>
+          <Typography variant="h1">All Reports</Typography>
+          {reports.map((report: FinancialSummaryRecord, index) => {
```
julianrkung: We should probably sort this descending (so the 2nd most recent Financial Report is shown, then 3rd, then 4th, etc)

src/screens/Reports/components/ReportCard.tsx
```diff
@@ -0,0 +1,64 @@
...
+export default function ReportCard(props: ReportCardProps): JSX.Element {
+  const { report, deadline, match } = props;
+  const classes = useStyles(props);
+
+  return (
+    <Card className={classes.root} variant="outlined">
+      <CardContent className={classes.content}>
+        {!props.current ?
+          <Typography color="textSecondary">{formatDateStringToLocal(report.lastUpdated)}</Typography>
+        : null}
```
julianrkung: NIT: you can do `{!props.current && <true-case>}` if your false case is `null`

src/lib/moment/momentUtils.ts
```diff
@@ -58,3 +59,13 @@ export const getCurrentMonthGracePeriodDeadline = () => {
   const gracePeriodDays : number = selectCurrentSiteGracePeriod(store.getState()) || 0;
   return moment().startOf('month').add(gracePeriodDays, 'days');
 }
+
+
+// Returns end of period plus number of days in grace period
+// Formats & converts to string
+export const getGracePeriodDeadline = (report: FinancialSummaryRecord, site: SiteRecord) => {
```
julianrkung: API makes more sense to take in a string period and a site record. This function is only really useful for `FinancialSummaryRecord` stuff, whereas if you pass in a `period : string` instead of the report it's useful in more cases.

src/App.tsx
```diff
@@ -73,19 +75,20 @@ function App() {
       <AuthenticatedRoute path={'/customers/customer/starting-meter-reading/edit'} component={EditStartingMeter} exact />
       <AuthenticatedRoute path={'/customers/customer/records'} component={CustomerRecords} exact />
       <AuthenticatedRoute path="/financial-summaries" component={FinancialSummariesMain} />
```
julianrkung: Can you delete the `/financial-summaries` endpoint since that's what `/reports` was supposed to be. 

It might be better to rename your current endpoint to `/financial-summaries` since it's the naming convention used in Airtable, but if you don't want to rename all your components you can just delete the `/financial-summaries` endpoint

src/screens/Reports/Report.tsx
```diff
@@ -0,0 +1,141 @@
...
+    ],
+    maintenance: [
+      { label: 'Completed', value: '0' },
+      { label: 'Incomplete', value: '0' },
+    ],
+    incidents: [
+      { label: 'Solved', value: '0' },
+      { label: 'Unsolved', value: '0' },
+    ],
+  }
```
julianrkung: It'd be easier to read if we didn't have this wrapping `info` object. We could just define the variables `profitInfo`, `customersInfo`, `inventoryInfo`, `maintenanceInfo`, and `incidentsInfo`, etc and pass them to `getContentsList` to render. 

I think we should delete the `info` object because the variable name is not very telling to the reader and it's hard to interpret at first glance because it's such a large object.

src/screens/Reports/Report.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+interface ReportProps extends RouteComponentProps {
+  classes: { section: string; header: string; list: string; icon: string; };
+  location: any;
+}
+
+interface Info {
+  label: string;
+  value: string;
+}
```
julianrkung: I'm not really in love with the one-use interface here since it seems like it's only applicable for the `getListContent` function (and won't be used outside that component / function). Its name is also really general and doesn't provide context (though I can't think of a better name off hand). 

We could accomplish the same thing by just adding a comment line to `getListContent` explaining what the caller should pass to it.

src/screens/Reports/components/ReportCard.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <Typography variant="h2">{report.period}</Typography>
+      </CardContent>
+      <CardActions className={classes.actions}>
+        <Link
+          to={{
+            pathname: `${match.url}/report`,
+            state: { report: report, deadline: deadline }
+          }}
+        >
+          <Button label={'View'} variant="outlined" />
```
julianrkung, line 57: As a note, we eventually need to translate any hardcoded text like "View". I think Micah hasn't worked his way through to the Financial Summary screen yet so he can probably take care of this one, but just wanted to get that on your radar for any future PRs.

In the future, you should look in `words.ts` to see if your hardcoded string is already available as something available to be translated and do internationalization around the text (you can look at `Inventory`-related stuff to see how to translate hardcoded strings)

src/screens/Reports/components/ReportCard.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <CardContent className={classes.content}>
+        {!props.current ?
+          <Typography color="textSecondary">{formatDateStringToLocal(report.lastUpdated)}</Typography>
+        : null}
+        <Typography variant="h2">{report.period}</Typography>
+      </CardContent>
+      <CardActions className={classes.actions}>
+        <Link
+          to={{
+            pathname: `${match.url}/report`,
```
julianrkung: We aren't really consistent with this (it'd be good to have all the `...Card` components route in the same way, but I'm down with hardcoding the path here to always be `/reports/report` so that you don't need to pass the `match` prop to the card (which I think feel a bit funky). I believe that `PurchaseRequestCard` does the hard-code method.

julianrkung (commented):
Only found tiny changes! I tested functionality and everything looked good so I'm down to merge after this iteration if nothing breaks :^)

src/screens/Reports/Report.tsx
```diff
@@ -0,0 +1,141 @@
...
+function Report(props: ReportProps) {
+  const { classes, match } = props;
+  const report = props.location.state.report;
+  const deadline = props.location.state.deadline;
+  const currentSite = useSelector(selectCurrentSiteInformation);
+
+  // Customers
+  const collected = `${round(report.totalAmountCollected)} Ks`;
+  const billed = `${round(report.totalAmountBilled)} Ks`;
+  const outstanding = `${round(report.totalAmountBilled - report.totalAmountCollected)} Ks`;
```
julianrkung: Wait actually I don't think this is how `outstanding` is calculated. On the Figma, this number is bigger than `Billed`. I'll ping ashley on slack and cc you

src/lib/redux/siteData.ts
cnnmon: Can't for same reason as selectCurrentSiteProfit

src/screens/Reports/Report.tsx
cnnmon: I want to keep it in to avoid any type - is there any way around this?

src/screens/Reports/Report.tsx
cnnmon: It's bigger and negative, though I thought that was just an error with how data was added

src/lib/redux/siteData.ts
```diff
@@ -35,3 +38,42 @@ export const updateTariffPlanInRedux = (tariffPlan: Partial<TariffPlanRecord>) =
...
+    totalUsage: 0,
+    totalAmountBilled,
+    totalAmountCollected,
+    totalAmountSpent: 0,
+    totalProfit,
+    inventoryAmountApproved,
+    inventoryAmountDenied,
+    period: getCurrentPeriod(),
+    isapproved: false,
+    lastUpdated: '0',
```
julianrkung: NIT: `lastUpdated` should be `formatDateToLocal(moment())` or whatever "right now" is. Not that we use it, but no point feeding bad data!

src/screens/Reports/FinancialSummaries.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+function FinancialSummaries(props: FinancialSummariesProps) {
+  const { classes, match } = props;
+  const currentSite = useSelector(selectCurrentSiteInformation);
+  const currentDeadline = formatDateStringToLocal(getCurrentMonthGracePeriodDeadline().toString());
+
+  const currentReport = selectCurrentPeriodFinancialSummary();
+  const reports: FinancialSummaryRecord[] = useSelector(selectAllFinancialSummariesArray) || [];
+  console.log("normal");
+  console.log(reports);
```
julianrkung: remove console logs before merging!

src/screens/Reports/FinancialSummaries.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  reports.sort((a: FinancialSummaryRecord, b: FinancialSummaryRecord) => {
+    const bDate = moment(b.lastUpdated);
+    const aDate = moment(a.lastUpdated);
+    const difference = bDate.diff(aDate, 'minutes')
+    return difference;
+  });
+
+  console.log("reverse");
+  console.log(reports);
```
julianrkung: remove console logs before merging!

src/screens/Home/Home.tsx
```diff
@@ -167,7 +166,7 @@ function Home(props: HomeProps) {
                 <Button label='Current Period' startIcon={<DescriptionIcon />} />
               </Link>
             </div>
-            <Link to={'/financial-summary'}>
+            <Link to={'/financial-summaries'}>
```
julianrkung, line 169: Both The way we pass `report` and the fact that we can't pass stuff through a `button` both don't really translate well to going immediately to the current financial summary.


I'm down to merge and figure this out in a subsequent PR since I think what we have now is good :^)

julianrkung (approved):
Could you also do the following cleanup before the merge:
1. delete the `FinancialSummary.tsx` file and 
2. remove the `/financial-summary` route in `App.tsx` 
3. Make the button "Current Period" button on the home screen not go to the `/financial-summary` route (it can just go to `/financial-summaries` for now) and put a TODO to "Make current period button navigate to current financial summary" 

Thank you so much for getting this done ;-; I'll get you all the weeb things you want as your bp gift HAHAHA

PR #140, opened 2021-04-24
Rounding Ks to 2 Decimal Places

### LINKED ISSUE #91: Floating point numbers aren't rounded to standardized significant figures
We should round all float logic. We should do this in a consistent manner throughout the app!

image

_Originally posted by @wangannie in https://github.com/calblueprint/meepanyar/pull/75#discussion_r609228254_

cnnmon:

# What's new in this PR

Rounds numbers on the Customer Main screen to two decimal points at most.
Migrated rounding function from https://github.com/calblueprint/meepanyar/pull/135.
resolves https://github.com/calblueprint/meepanyar/issues/91

## How to review

Customer Main.

### Screenshots

image

CC: @julianrkung

### DIFF
--- a/src/lib/redux/siteData.ts
+++ b/src/lib/redux/siteData.ts
@@ -15,7 +15,7 @@ export const selectCurrentSiteInformation = createSelector(
   (siteId, state) => state.siteData.sites[siteId].siteInformation)
 
 export const selectCurrentSiteGracePeriod = createSelector(
-  selectCurrentSiteInformation, 
+  selectCurrentSiteInformation,
   (currentSiteRecord : SiteRecord) => currentSiteRecord.gracePeriod)
 
 
@@ -34,4 +34,10 @@ export const updateTariffPlanInRedux = (tariffPlan: Partial<TariffPlanRecord>) =
 
 export const updateSiteInRedux = (siteUpdates : Partial<SiteRecord>) => {
   store.dispatch(updateSite(siteUpdates))
+}
+
+// Rounds number to at most 2 decimal points
+// And converts to string
+export const round = (number: number) => {
+  return (Math.round(number * 100) / 100).toString();
 }
--- a/src/lib/redux/siteData.ts
+++ b/src/lib/redux/siteData.ts
@@ -35,9 +35,3 @@ export const updateTariffPlanInRedux = (tariffPlan: Partial<TariffPlanRecord>) =
 export const updateSiteInRedux = (siteUpdates : Partial<SiteRecord>) => {
   store.dispatch(updateSite(siteUpdates))
 }
-
-// Rounds number to at most 2 decimal points
-// And converts to string
-export const round = (number: number) => {
-  return (Math.round(number * 100) / 100).toString();
-}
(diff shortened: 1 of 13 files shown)


src/lib/redux/siteData.ts
```diff
@@ -34,4 +34,10 @@ export const updateTariffPlanInRedux = (tariffPlan: Partial<TariffPlanRecord>) =
 
 export const updateSiteInRedux = (siteUpdates : Partial<SiteRecord>) => {
   store.dispatch(updateSite(siteUpdates))
-}
\ No newline at end of file
+}
+
+// Rounds number to at most 2 decimal points
+// And converts to string
+export const round = (number: number) => {
```
julianrkung: Can we actually use this function for all of the numbers we display? We should uniformally use this function (maybe it'd be good to place this in another place like put it in a new `utils` file.


we almost might just be able to use this https://www.w3schools.com/jsref/jsref_tofixed.asp `toFixed` function instead of creating a new function (we might have to watch out for number vs string types though)

edit: I'm fine with the existing function since it converts to at msot 2 decimal points.

julianrkung (commented):
We don't need to do this for forms (we can do that in a later PR using Formik), but we should do this for all the numbers we display on both Customer and Inventory (and eventually FInancial Summaries).

I'll try to make a list, but double check if I miss anything
1. Remaining Balance in `CustomerProfile`
2. "Last Recorded Reading" and "Starting Meter"
3. Fixed Tariff, Unit Tariff, Free Tariff in `/Profile` screens and `CustomerProfile`
4. Ending  Meter, Period Usage, Amount Billed in `CustomerProfile`
5. `Remaining Balance` in Payment screen
6. `Starting Reading (aka Current Reading)` in Meter Reading screen
7. Meter Reading / Payment amounts in `/records` screens
8. Inventory amounts in `InventoryMain`
9. "Date of Purchase" and  Ks Amounts in `All Purchases` screen
10. All numbers in `InventoryProfile` screen
11. The amount in `Customer Updates` screen
12. The amount in `Inventory Purchase` screen
13. "Amount Spent" and "Amount Purchased" in `PurchaseRequest` screen

src/lib/redux/siteData.ts
wangannie: +1 on moving this to a new more general utils file

wangannie: But if you do stick with this helper rather than `toFixed`, I don't think it's immediately intuitive that `round` returns a string -- maybe rename to something like `roundToString`?

src/screens/Customers/AddMeterReading.tsx
```diff
@@ -87,7 +87,7 @@ function AddMeterReading(props: AddMeterReadingProps) {
   }
 
   const cardInfo = [{
-    number: startingMeterAmount.toString(),
+    number: round(startingMeterAmount),
```
wangannie: Nit: I guess it was always like this but it feels rather unintuitive that the `number` prop is actually a `string` type. Not totally necessary but I feel like it would be more intuitive to rename to something like `value`.

wangannie (approved):
I poked around all over the app and I didn't catch any stray un-rounded decimals! Nice work 🪂 ! I just let some nitpicky comments that aren't super essential, so feel free to address or ignore.

Also, make sure you **rename this PR** before you merge since this now covers the entire app, not just Customer.

